PR #11, opened 2020-01-17
Undefined name: import operator

cclauss:
__import [operator](https://docs.python.org/3/library/operator.html)__ to avoid raising NameError at runtime.

### DIFF
--- a/crosshair/objectproxy.py
+++ b/crosshair/objectproxy.py
@@ -2,11 +2,6 @@ import copy
 import operator
 import sys
 
-try:
-    long
-except NameError:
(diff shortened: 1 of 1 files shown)


crosshair/objectproxy.py
```diff
@@ -2,6 +2,11 @@
 import operator
 import sys
 
+try:
+    long
+except NameError:
+    long = int
```
pschanely: "long" is later used in the definition for the `__long__` method, which I think can also be removed. Would you mind removing that method too?

Issue #3, opened 2019-10-19
Implement a symbolic collections.deque


pschanely:
Per #2, deque cannot be analyzed symbolically. Tentative plan is to implement a "dumb" version on top of a generic list: https://github.com/pschanely/CrossHair/blob/56ce14ee44c1103121c483f479d3d1a5a4bd3438/crosshair/simplestructs.py#L57

oneEdoubleD:
I'll take a look at this once you've finished refactoring the implementation pattern

pschanely:
Thank you for taking a look! I've given the basic framework for how I imagined it here:

https://github.com/pschanely/CrossHair/commit/7a4e4606fdbbb6c34825ae79e5c25af4ca5ff528

But use your best judgement too, and feel free to be noisy in the chat while you investigate!

pschanely:
Just a note for later: pypy actually also has a pure python implementation of this: https://bitbucket.org/pypy/pypy/src/default/lib_pypy/_collections.py

amacfie:
> Just a note for later: pypy actually also has a pure python implementation of this: https://bitbucket.org/pypy/pypy/src/default/lib_pypy/_collections.py

@pschanely If we run CrossHair with pypy does that remove the need for symbolic versions of stuff pypy implements in pure python like `collections.deque`? Are there other benefits of running CrossHair with pypy?

PR #13, opened 2020-01-18
Implemented collections deque

oneEdoubleD:
I've implemented the symbolic collections.deque class along with a few test cases and the change to `make_registrations()`. Let me know if there's anything else to do, or if you'd like any other test cases covered!

### DIFF
--- a/crosshair/libimpl/collectionslib.py
+++ b/crosshair/libimpl/collectionslib.py
@@ -67,15 +67,9 @@ class ListBasedDeque:
         self._contents.reverse()
 
     def rotate(self, n: Optional[int]=1) -> None:
-        if n == 0:
-            # No rotation
-            pass
-        if n > 0:
-            # Rotate to the right
-            self._contents = self._contents[-n:] + self._contents[:-n]
-        if n < 0:
-            # Rotate to the left
-            self._contents = self._contents[n:] + self._contents[:n]
+            if not self._contents or n % len(self._contents) == 0:
+                pass
+            self._contents = self._contents[-n % len(self._contents):] + self._contents[:-n % len(self._contents)]
 
     def maxlen(self) -> int:
         return self._maxlen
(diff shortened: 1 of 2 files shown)


oneEdoubleD:
This is related to https://github.com/pschanely/CrossHair/issues/3

crosshair/libimpl/collectionslib_test.py
```diff
@@ -12,28 +13,102 @@
...
-            pre: len(l) > 0
-            post: len(l) != 222
-            '''
-            return l
-        # Right now, this returns unknown rather than fail, because creation of
-        # the deque forces the list length to become concrete.
-        self.assertEqual(*check_fail(f))
+
+    def setUp(self):
+        self.test_list = ListBasedDeque([1, 2, 3, 4, 5])
```
pschanely: Personal preference thing (so you should feel free to ignore if you like): I tend to bias towards repeating the test data in each test. It adds an extra line to a lot of tests, but makes them easier to read.

crosshair/libimpl/collectionslib.py
```diff
@@ -8,14 +8,71 @@ class ListBasedDeque:
     def __init__(self, contents: List[T], maxlen: Optional[int]=None):
```
pschanely: Making the class extend `collections.abc.Sequence` is possibly a good idea?
Which makes me also think that some list-like deque behavior might not work properly yet?:
```
>>> from collections import deque
>>> 3 in deque([3,4])
True
>>> deque([3,4]) < deque([4,3])
True
>>> deque([3,4]) > deque([4,3])
False
>>> deque([3,4]) == deque([3,4])
True
```

crosshair/libimpl/collectionslib_test.py
```diff
@@ -12,28 +13,102 @@
...
+            self.test_list.index(1, start=2)
+
+        self.assertTrue('1 is not in list' in str(context.exception))
+    
+    def test_deque_index_with_start_and_end_index(self) -> None:
+        i = self.test_list.index(2, start=0, end=3)
+        self.assertEqual(i, 1)
+    
+    def test_deque_index_with_start_and_end_index_throws_correct_exception(self) -> None:
+        with self.assertRaises(ValueError) as context:
```
pschanely: Thanks for exception checking!

pschanely:
I'll merge what we've got so far. Let's continue to work on it, though!

Issue #39, opened 2021-01-04
Symbolic string method support


pschanely:
Today, many string methods cause symbolic strings to be "materialized" (the symbolic string becomes a concrete string and then gets handed to the native string implementation). 

For some of these operations, this is likely the best we can do, but others may be implementable with the SMT solver directly. This is a list of those methods, and we'll check them off when we've decided that we have the implementation we want. For each method, we'll decide to:
1. Leave as-is, with materialization in place.
2. Implement the method symbolically, using Z3.
3. Implement the method in terms of other string methods that have symbolic implementations.

Note further that direct unicode support in Z3 isn't yet implemented. We might want to defer spending time on this until that's resolved or until we implement a workaround (as a sequence of bitvectors ... or a massive enum?)

- [x]   def capitalize(self)
- [x]   def casefold(self)
- [x]   def center(self, width, *args)
- [x]   def count(self, sub, start=0, end=sys.maxsize)
- [x]   def encode(self, encoding=_MISSING, errors=_MISSING)
- [x]   def endswith(self, suffix, start=0, end=sys.maxsize)
- [x]   def expandtabs(self, tabsize=8)
- [x]   def find(self, sub, start=0, end=sys.maxsize)
- [x]   def format(self, *args, **kwds)
- [x]   def format_map(self, mapping)
- [x]   def index(self, sub, start=0, end=sys.maxsize)
- [x]   def isalpha(self)
- [x]   def isalnum(self)
- [x]   def isascii(self)
- [x]   def isdecimal(self)
- [x]   def isdigit(self)
- [x]   def isidentifier(self)
- [x]   def islower(self)
- [x]   def isnumeric(self)
- [x]   def isprintable(self)
- [x]   def isspace(self)
- [x]   def istitle(self)
- [x]   def isupper(self)
- [x]   def join(self, seq)
- [x]   def ljust(self, width, *args)
- [x]   def lower(self): return self.data.lower()
- [x]   def lstrip(self, chars=None)
- [x]   def partition(self, sep)
- [x]   def replace(self, old, new, maxsplit=-1)
- [x]   def rfind(self, sub, start=0, end=sys.maxsize)
- [x]   def rindex(self, sub, start=0, end=sys.maxsize)
- [x]   def rjust(self, width, *args)
- [x]   def rpartition(self, sep)
- [x]   def rsplit(self, sep=None, maxsplit=-1)
- [x]   def rstrip(self, chars=None)
[...]

PR #42, opened 2021-01-08
WIP: Implement zfill

davidmrdavid:
Related to: https://github.com/pschanely/CrossHair/issues/39
Previously, using CrossHair on a function with `zfill`  required Z3 to use a "best effort" approach: materialize concrete values of strings as inputs and try its best to find a counter example to the user-provided post-condition. This PR attempts to provide a symbolic implementation of `zfill` using pre-existing functionality.

So far, this is passing a simple unit test! More testing is needed, especially around negative width values. _This is just a start!_

**Heads up:**
At the time of writing, this is a very naive approach. I expect we'll go through a few iterations of this.
I also left some questions as comments on my implementation, would love to hear your thoughts.
Finally, I realize this is a very naive first PR. I promise future ones will be a bit more substantial once I get more familiarized with the design of CrossHair :) 

Thanks!

### DIFF
--- a/crosshair/libimpl/builtinslib.py
+++ b/crosshair/libimpl/builtinslib.py
@@ -1678,10 +1678,11 @@ class SmtStr(AtomicSmtValue, SmtSequence, AbcString):
         return ret
     
     def zfill(self, width: int):
-        size = SmtInt(z3.Length(self.var)) # Why does this need to be casted to SmtInt?
-        offset = width - size # Unsure about the type of offset at this point
-        buffer = SmtStr("0") * offset
-        return buffer + self
+        size = len(self)
+        offset = width - size
+        buffer = "0" * offset
+        filled_str = buffer + self
+        return filled_str
 
 _CACHED_TYPE_ENUMS: Dict[FrozenSet[type], z3.SortRef] = {}
 
--- a/crosshair/libimpl/builtinslib.py
+++ b/crosshair/libimpl/builtinslib.py
@@ -1676,7 +1676,12 @@ class SmtStr(AtomicSmtValue, SmtSequence, AbcString):
         new_maxsplit = -1 if maxsplit == -1 else maxsplit - 1
         ret.extend(self[first_occurance+1:].split(sep=sep, maxsplit=new_maxsplit))
         return ret
-
+    
+    def zfill(self, width: int):
+        size = SmtInt(z3.Length(self.var)) # Why does this need to be casted to SmtInt?
+        offset = width - size # Unsure about the type of offset at this point
+        buffer = SmtStr("0") * offset
+        return buffer + self
 
 _CACHED_TYPE_ENUMS: Dict[FrozenSet[type], z3.SortRef] = {}
 
@@ -2014,8 +2019,7 @@ def make_registrations():
             'splitlines',
             'startswith',
             'strip',
-            'translate',
-            'zfill',
+            'translate'
     ]:
         orig_impl = getattr(orig_builtins.str, name)
         register_patch(orig_builtins.str, with_realized_args(orig_impl), name)
--- a/crosshair/libimpl/builtinslib_test.py
+++ b/crosshair/libimpl/builtinslib_test.py
@@ -289,6 +289,12 @@ class NumbersTest(unittest.TestCase):
 
 class StringsTest(unittest.TestCase):
 
+    def test_zfill_fail(self) -> None:
+        def f(s: str) -> str:
+            ''' post: __return__ != "00012"'''
+            return s.zfill(5)
+        self.assertEqual(*check_fail(f))
+
     def test_cast_to_bool_fail(self) -> None:
         def f(a: str) -> str:
             ''' post: a '''
--- a/func.py
+++ b/func.py
@@ -1,3 +1,0 @@
-def find_zfill(s: str) -> str:
-  ''' post: __return__ != "00012" '''
-  return s.zfill(5)
--- a/func.py
+++ b/func.py
@@ -1,0 +1,3 @@
+def find_zfill(s: str) -> str:
+  ''' post: __return__ != "00012" '''
+  return s.zfill(5)


crosshair/libimpl/builtinslib.py
```diff
@@ -1676,7 +1676,12 @@ def split(self, sep:Optional[str] = None, maxsplit:int = -1):
         new_maxsplit = -1 if maxsplit == -1 else maxsplit - 1
         ret.extend(self[first_occurance+1:].split(sep=sep, maxsplit=new_maxsplit))
         return ret
-
+    
+    def zfill(self, width: int):
+        size = SmtInt(z3.Length(self.var)) # Why does this need to be casted to SmtInt?
```
pschanely, line 1681: Great question! Essentially:

For each Python type (say, `int`), there are two important but different other types of values that CrossHair cares about: (1) a type that emulates the built-in type (`SmtInt`) and, (2), a z3 sort (`z3.IntSort()`). The Smt* type tries very hard to be indistinguishable from the python type, while the z3 expression is just a regular z3 expression.

When crosshair analyzes a function that takes, say, an `int`, we pass in a value of type `SmtInt`. Accessing the `.var` attribute of `SmtInt` gets us to a z3 expression of sort `z3.IntSort()`. To go the other way (from z3 expression to the `SmtInt` value), we construct a new `SmtInt`, passing in a z3 expression. (it's actually a constructor, not a cast)

So, it's common to see patterns where we access `.var` on some Smt* objects, then do some work in pure z3, and then wrap that result up in some more Smt* objects.

In this case, you can fall back to some other work we've done that already interprets `len()` symbolically: this line is the same as just saying "size = len(sellf)" (and I'd probably recommend that here). For fun, look at `SmtSequence.__len__` - it is exactly your line here.

crosshair/libimpl/builtinslib.py
```diff
@@ -1676,7 +1676,12 @@ def split(self, sep:Optional[str] = None, maxsplit:int = -1):
...
+        offset = width - size # Unsure about the type of offset at this point
```
pschanely, line 1682: `size` has type `SmtInt`. `width` could be either a real `int` or another `SmtInt`. Either way, the subtraction operator is overridden and will produce an `SmtInt` as a result!

crosshair/libimpl/builtinslib.py
```diff
@@ -1676,7 +1676,12 @@ def split(self, sep:Optional[str] = None, maxsplit:int = -1):
...
+        buffer = SmtStr("0") * offset
```
pschanely, line 1683: Ah, this is a gotcha; the constructor of all Smt* types take a first argument of type `Union[str, z3.ExprRef]`.

When you pass a string, you are naming a fresh z3 variable (all z3 variables must have names). This is how we create, say, an unconstrained  integer to pass into a user function.

That distinction is super confusing here. `SmtStr("0")` is actually creating a new symbolic string with an arbitrary, unknown value and having the **name** "0".  Instead, you could do `SmtStr(z3.StringVal("0"))`.

That said, the `*` operator is overridden by SmtInt (via `__rmul__`) and already does the symbolic work you need ([check it out here, for fun](https://github.com/pschanely/CrossHair/blob/94df30a5a4dd3a7956b34909dfc2ce9f2179eb71/crosshair/libimpl/builtinslib.py#L616-L626)); you might be able to just say `buffer = "0" * offset`.

Taking all my suggestions together, you may be surprised, saying "wait, I'm just writing an implementation of zfill in regular old python - I'm not using z3 at all!" Indeed, a lot of the CrossHair codebase is just re-implementing the python library in regular old python (instead of the C-based implementation that it ships with). We just take special care to ensure we're always building on top of python that we already symbolically handle. This is a good way to start, but I'm happy to help you find something next that will require digging into the solver!

davidmrdavid: Thanks for all the detailed explanation, to this and my other questions :) 🎉 

I'm glad to hear that so many primitive operations are already mapped to their symbolical equivalents. Indeed, it's kinda magical to implement more functionality in regular "good old python" have z3 understand how to test that so quickly.

One follow-up question about `__mul__`, found [here](https://github.com/pschanely/CrossHair/blob/94df30a5a4dd3a7956b34909dfc2ce9f2179eb71/crosshair/libimpl/builtinslib.py#L616-L626).

The `self.statespace` variable caught my eye. Is that _just_ the space / conjunction of constraints? It would appear so, but I wanted to make sure. It would also appear that the statespace is saved in thread-local storage, which I found noteworthy: https://github.com/pschanely/CrossHair/blob/94df30a5a4dd3a7956b34909dfc2ce9f2179eb71/crosshair/statespace.py#L127-L130
Any reason why?

I hope these are not too bothersome to answer. If so, feel free to just leave it at "you'll understand later" and we can move on :) . Thanks again!

davidmrdavid:
Thanks again for the feedback and pointers! Thinking of adding a few more tests and maybe some comments to wrap this up. As for next steps after that, I'm still interested in tackling a few of the other items in this ticket: https://github.com/pschanely/CrossHair/issues/39 .

I'm thinking of tackling `capitalize` next, which sounds easy-enough in theory but I'm not too certain how to map it to constraints "the CrossHair way". Does that sound like a reasonable next task to tackle? If so, I'll start hacking on it! I do have some ideas on how to start, though it might not be pretty :P

crosshair/libimpl/builtinslib.py
pschanely: Please continue to ask questions! I'm hoping to steal from these responses to build a developer's guide :)

Maybe a preliminary that might or might not be obvious: CrossHair doesn't do any kind of AST analysis or bytecode manipulation. It just repeatedly calls the function you want to analyze and passes in special objects that are made to look like the types that the function actually takes. When stuff happens to those special objects, we consult the SMT solver and then produce other special objects, and the function continues.

The statespace is a thread local that's created for each execution of the analyzed function; it serves a few purposes that are central to how CrossHair works, but the most important two are:
(1) It stores the current set of z3 constraints for this execution. (it holds a `z3.Solver`)
(2) It consults and updates an execution branch tree. This tree represents the different execution paths that the analyzed function could take. It uses this tree to avoid taking branches that it's already explored on prior executions and updates the tree with the novel parts of its own execution.

The statepsace almost doesn't need to be a thread-local (and wasn't until recently), but there is a case or two where it's challenging to get at it. One problem is that since CrossHair actually calls the function under analysis, you may find yourself in part of the stack where you can't easily get at the statespace. (imagine that you need to invent a new symbolic value out of nowhere, for instance) Furthermore, it would always be an error to work with values from different statespaces, so we aren't really losing a lot of flexibility by making it thread-local.

pschanely:
> Thanks again for the feedback and pointers! Thinking of adding a few more tests and maybe some comments to wrap this up. As for next steps after that, I'm still interested in tackling a few of the other items in this ticket: #39 .

Fantastic, and thank you! I'd add a test that does `check_ok()`; I'm curious which of the [let's prove leftpad](https://github.com/hwayne/lets-prove-leftpad#why-are-we-proving-leftpad) properties we can prove. You might find it interesting to consider which ones are (not) provable and why. (I don't even know what you'll encounter!)

> I'm thinking of tackling `capitalize` next, which sounds easy-enough in theory but I'm not too certain how to map it to constraints "the CrossHair way". Does that sound like a reasonable next task to tackle? If so, I'll start hacking on it! I do have some ideas on how to start, though it might not be pretty :P

That sounds like a good one to try next. It'll be challenging, but will probably force you to hit a lot of important issues. And may require that you do something with the solver. I'd start by thinking first about how to do this in pure z3 ... that is, how can I express that one z3 string is the capitalized form of a different z3 string? (z3 doesn't natively understand uppercase/lowercase!) I just enabled GitHub "discussions" which might be a good place to start a discussion about how to approach this.

crosshair/libimpl/builtinslib.py
```diff
@@ -2014,8 +2020,7 @@ def make_registrations():
             'splitlines',
             'startswith',
             'strip',
-            'translate',
-            'zfill',
```
pschanely, line 2018: OOC, how did you end up with this edit? I haven't explained this section yet, but it exists to handle the case where a **native** python string calls `zfill` on a **symbolic** argument (SmtInt). (we do this by monkey-patching the built-in str class!!)

It might be ok to remove, as we've got some fallback tactics to detect this kind of case, but if you don't need this change, I'd defer it for another CL (or wait for me to pick it up!)

davidmrdavid: Ooh, interesting. I came up with that change simply because I did a CTRL+F for `zfill` and assumed this section of code was providing a fallback for `zfill` **in general**, not just for the case you outline (native string calls zfill with SMTInt).

In any case, I'll take a mental note to remove this change and to investigate this part of the code in the future!

pschanely:
Sadly, I just noticed that the zfill() function is more complex than I'd expected! If the given string starts with a `+` or `-`, that character stays at the end after filling. ([docs](https://docs.python.org/3/library/stdtypes.html#str.zfill))

Again, I think we can just implement this logic in plain Python to handle such cases.

Frankly, I'm not sure how often zfill is used in practice. Your investigation into lower/uppercase is probably much more impactful, so you might want to prioritize that. 😄

Rik-de-Kort:
I think we can close this, since `zfill` functionality was implemented in #95 :)

PR #43, opened 2021-01-19
Add necessary import of type to example

JulianWgs:
Less friction when copy and pasting the example


pschanely (approved):
Indeed! Thank you!

If you try this out on real examples, I'd love to hear how that goes!

PR #52, opened 2021-01-28
Run CrossHair as module

mristin:
This patch allows CrossHair to be run as a module, *i.e.*,
`python -m crosshair ...`. This is necessary for the environments which
depend on Python interpreter (*e.g.*, IDEs in Windows).


pschanely:
Thanks for this! I certainly would not have caught the detail about the prog name!

PR #65, opened 2021-02-04
Reformat code with black 20.8b1

mristin:
This is an initial reformatting of the code. It is intentionally kept
separate from the commit where the pre-commit script is introduced so
that: 1) we can inspect the formatting in isolation, and 2) we can
review precommit checks in isolation.

### DIFF
--- a/crosshair/core_test.py
+++ b/crosshair/core_test.py
@@ -24,49 +24,45 @@ from crosshair.util import debug
 from crosshair.util import set_debug
 
 
-
-
-
-
-
-
 #
 # Begin fixed line number area.
 # Tests depend on the line number of the following section.
 #
 
+
 class Pokeable:
-    '''
+    """
     inv: self.x >= 0
-    '''
+    """
+
     x: int = 1
 
     def poke(self) -> None:
-        '''
+        """
         post[self]: True
-        '''
+        """
         self.x += 1
 
     def wild_pokeby(self, amount: int) -> None:
-        '''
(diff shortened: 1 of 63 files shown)


mristin:
The line length of black is set to default 88 characters. This worked pretty well.

I also had to change the test files to conform with the reformatted code.

pschanely:
I do find single single-quotes easier on the hands, but this'll autoformat eventually right? So I can just keep with my curmudgeonly ways? :)
Looking really good overall.

crosshair/core_test.py
```diff
@@ -24,49 +24,45 @@
...
 
 
-
-
-
-
-
-
 #
 # Begin fixed line number area.
```
pschanely, line 28: Right, so this "fixed line number area" is an odd/hacky case - I have these spare newlines here so that I can just cut out some when import lines are added without changing the tests.

I imagine we could figure out a different way to handle this though. Maybe base the line numbers in the tests as offsets from the line numbers derivable from the function objects themselves.

mristin: I think there were so few breaking tests that I would suggest just to update them manually at the moment?

mristin:
> I do find single single-quotes easier on the hands, but this'll autoformat eventually right? So I can just keep with my curmudgeonly ways? :)
> Looking really good overall.

Yes, black takes care of that :).

pschanely:
So cool; thank you!

PR #70, opened 2021-02-04
Fix expected coverage in `util_test`

mristin:
The previous threshold for opcode coverage was set to 0.9. However, on
my Windows machine, the coverage is 0.88.

This patch lowers the threshold arbitrarily to 0.85.


pschanely:
Fascinating!

PR #71, opened 2021-02-04
Add precommit script

mristin:
This patch introduces a commit script to run one, more or all pre-commit
checks from the command line or as part of our continuous integration
pipeline.

It also documents the contributing guidelines, which also explain how to
set up the development environment.

### DIFF
--- a/precommit.py
+++ b/precommit.py
@@ -1,0 +1,142 @@
+#!/usr/bin/env python3
+"""Run pre-commit checks on the repository."""
+import argparse
+import enum
+import os
+import pathlib
+import subprocess
+import sys
+
+
+class Step(enum.Enum):
+    BLACK = "black"
+    FLAKE8 = "flake8"
+    PYDOCSTYLE = "pydocstyle"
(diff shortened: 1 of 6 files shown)


precommit.py
```diff
@@ -0,0 +1,142 @@
...
+
+        # fmt: off
+        subprocess.check_call(
+            [
+                "coverage", "run",
+                "--source", "crosshair",
+                "--omit=__init__.py"
+                "--omit=*_test.py",
+                "-m", "pytest",
+                "--doctest-modules"
```
pschanely, line 119: I **think** this will run the docstring doctests and potentially we could add  `--doctest-glob="*.md"` to cover the doctest target below.

I'll merge and you can follow up if you like.

mristin: Actually, I did the separation on purpose since the tests take quite long to run. I'll add a comment in the script.

pschanely: SGTM. The doctest coverage right now doesn't make for a great "smoke test," but we'll work on it!

PR #74, opened 2021-02-04
Publish continuously to PyPI

mristin:
This patch introduces a GitHub workflow that is triggered on a GitHub
release. The workflow automatically builds the source distribution and
uploads it to PyPI.

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -1,0 +1,36 @@
+# Based on https://github.com/actions/starter-workflows/blob/main/ci/python-publish.yml
+name: Publish to Pypi
+
+on:
+  release:
+    types: [created]
+
+  push:
+    branches:
+      - '*/fixed-publishing-to-pypi'
+
+jobs:
+  deploy:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install setuptools wheel twine
+
+    - name: Build and publish
+      env:
+        TWINE_USERNAME: "__token__"
+        TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
+      run: |
+        python setup.py bdist_wheel sdist
+        twine upload dist/*
--- a/setup.py
+++ b/setup.py
@@ -34,6 +34,7 @@ setup(
             "codecov",
             "black==20.8b1",
             "pydocstyle==5.1.1",
+            "wheel",
         ]
     },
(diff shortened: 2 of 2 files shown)


mristin:
@pschanely you need to create a PyPI token for this repository (see [this section in PyPI help](https://pypi.org/help/#apitoken)) and you need to set it as `PYPI_TOKEN` GitHub secret (see [this StackOverflow answer](https://stackoverflow.com/a/57706400/1600678)).

.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,36 @@
...
+      run: |
+        python -m pip install --upgrade pip
+        pip install setuptools wheel twine
+
+    - name: Build and publish
+      env:
+        TWINE_USERNAME: "__token__"
+        TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
+      run: |
+        python setup.py sdist
```
pschanely, line 35: Personally, I've been doing `python setup.py bdist_wheel sdist` which gives us a wheel too. I don't really know what the practical difference is.

As you can tell, I've only been putting releases up on PyPI, not labeling/branching anything in github, so I might need your help with that too :)

pschanely:
Secret should be in place now.

.github/workflows/publish-to-pypi.yml
mristin: Aren't wheels useful only when you have compiled binaries included in the package?

Re releases: here is the [relevant GitHub doc page](https://docs.github.com/en/github/administering-a-repository/managing-releases-in-a-repository).

I usually create a separate commit (such as "Release 1.2.3") which contains the change log in the message body (as bullet list). Then I create the release with the tag "v1.2.3" and the title of the release is "CrossHair 1.2.3". I repeat the changelog in the release description to make it easier for the users to find it. And that's it :).

pschanely: Thanks!
And, yeah, maybe the wheel [installs a little bit faster](https://stackoverflow.com/questions/59619940/what-is-the-point-of-built-distributions-for-pure-python-packages)? Low stakes either way I suppose.

mristin: Oh, I didn't know that, thanks! Let me add bdist then before you merge.

mristin:
@pschanely  I forced a commit amended with `bdist_wheel`; could you please re-review?

setup.py
```diff
@@ -34,6 +34,7 @@
             "codecov",
             "black==20.8b1",
             "pydocstyle==5.1.1",
+            "wheel",
```
mristin, line 37: I added wheel locally so that the developers can build on their machines, if necessary. Otherwise, there is an error message (see [this StackOverflow question](https://stackoverflow.com/questions/34819221/why-is-python-setup-py-saying-invalid-command-bdist-wheel-on-travis-ci)).

PR #75, opened 2021-02-05
Undry metadata from setup.py

mristin:
We need to denormalize the medata (such as version, author, license
*etc.*) between `setup.py` and `crosshair/__init__.py` so that both
Sphinx and setup can access them.

Mind that you can not import the module (`crosshair/__init__.py`) from
within `setup.py` as this would cause race condition during the
installation.

To enforce the consistency, we add a script to the pre-commit checks.

### DIFF
--- a/check_init_and_setup_coincide.py
+++ b/check_init_and_setup_coincide.py
@@ -1,0 +1,122 @@
+#!/usr/bin/env python3
+
+"""Check that the distribution and crosshair/__init__.py are in sync."""
+import subprocess
+import sys
+from typing import Optional, Dict
+
+import pkg_resources
+import email
+
+import crosshair
+
+
+def main() -> int:
+    """Execute the main routine."""
+    success = True
+
+    ##
+    # Check basic fields
+    ##
+
+    setup_py = dict()  # type: Dict[str, str]
+
+    fields = ["version", "author", "license", "description"]
+    for field in fields:
+        out = subprocess.check_output(
+            [sys.executable, "setup.py", f"--{field}"], encoding="utf-8"
+        ).strip()
+
+        setup_py[field] = out
+
+    if setup_py["version"] != crosshair.__version__:
+        print(
+            f"The version in the setup.py is {setup_py['version']}, "
+            f"while the version in crosshair/__init__.py is: "
+            f"{crosshair.__version__}",
+            file=sys.stderr,
+        )
+        success = False
+
+    if setup_py["author"] != crosshair.__author__:
+        print(
+            f"The author in the setup.py is {setup_py['author']}, "
+            f"while the author in crosshair/__init__.py is: "
+            f"{crosshair.__author__}",
+            file=sys.stderr,
+        )
+        success = False
+
+    if setup_py["license"] != crosshair.__license__:
+        print(
+            f"The license in the setup.py is {setup_py['license']}, "
+            f"while the license in crosshair/__init__.py is: "
+            f"{crosshair.__license__}",
+            file=sys.stderr,
+        )
+        success = False
+
+    if setup_py["description"] != crosshair.__doc__:
+        print(
+            f"The description in the setup.py is {setup_py['description']}, "
+            f"while the description in crosshair/__init__.py is: "
+            f"{crosshair.__doc__}",
+            file=sys.stderr,
+        )
+        success = False
+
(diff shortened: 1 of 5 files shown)


crosshair/__init__.py
```diff
@@ -1,7 +1,16 @@
+"""Statically analyze Python code using symbolic execution."""
```
mristin, line 1: I had to change the description to conform with PEP 257.

crosshair/__init__.py
```diff
@@ -1,7 +1,16 @@
...
 from crosshair.core import register_patch
 from crosshair.core import register_type
 from crosshair.statespace import StateSpace
 from crosshair.util import IgnoreAttempt
 from crosshair.util import debug
+
+# Do not forget to update in setup.py!
+__version__ = "0.0.9"
+__author__ = "Phillip Schanely"
+__copyright__ = "Copyright 2019-2021 Phillip Schanely"
```
mristin, line 14: Please double-check the copyright! It will also be used in Sphinx in the following pull request.

crosshair/__init__.py
pschanely: FYI, I've been shying away from the "static analysis" and "dynamic analysis" text in various places. The problem with CrossHair is that people familiar with either often want to say that CrossHair is the other thing.
I do intend it to be used like you'd use a static analysis tool, though, so feel free to assert this. :)

check_init_and_setup_coincide.py
```diff
@@ -0,0 +1,122 @@
...
+import sys
+from typing import Optional, Dict
+
+import pkg_resources
+import email
+
+import crosshair
+
+
+def main() -> int:
```
pschanely, line 14: How would you feel about a dedicated metadata file in the package that setup.py can manually read/eval (e.g. that SO link you found has a suggestion along the lines of `open('crosshair/_metadata.py')`)

A little reluctant to suggest it, since you've put some amount of effort into this PR; but I just felt like the test was a fair amount of code to add, and if we can keep the metadata DRY, then we'll have a tiny bit less work to do at release time. WDYT?

crosshair/__init__.py
pschanely: Hmm. My vague impression is that by default copyright is held per-contribution. Can we leave it out, or say "varies" or something?

mristin:
(For some reason GitHub does not allow me to reply to the comment above, so I am replying here.)

> FYI, I've been shying away from the "static analysis" and "dynamic analysis" text in various places. The problem with CrossHair is that people familiar with either often want to say that CrossHair is the other thing.
I do intend it to be used like you'd use a static analysis tool, though, so feel free to assert this. :)

What about just "Analyze Python code for correctness using symbolic execution."?

check_init_and_setup_coincide.py
mristin: I tried the approach with the metadata in other projects (usually called `{project}_meta.py` like `icontract_meta.py`) and it worked for some time, but eventually the problems came around (either at the setup time, at the documentation build time, or on readthedocs servers). 

In the end, I had to denormalize after giving up the fight with setup.py, Sphinx, pip and readthedocs. I agree that it is clumsy to have such a long script test something so trivial, but at the moment that is the only way I know that works robustly.

mristin: (See [this meta file](https://github.com/Parquery/gs-wrap/blob/master/gswrap_meta.py), for example. This worked, but the very same approach did not work in icontract on readthedocs.)

mristin:
(Regarding the copyright)

> Hmm. My vague impression is that by default copyright is held per-contribution. Can we leave it out, or say "varies" or something?

You are correct. I'll remove it and try to fix Sphinx to ignore it.

mristin:
@pschanely I removed the copyright in a separate commit to make it easier for the re-review. Please squash and merge (instead of just merge).

pschanely:
I'm also having trouble adding per-thread comments!

> "Analyze Python code for correctness using symbolic execution"
SGTM!

> I'll remove it and try to fix Sphinx to ignore it.
Thanks! I'll squash+merge when you have confirmed it'll work - LMK!

mristin:
@pschanely amended. Please have another quick look to make sure I didn't forget something, but it should be ready.

PR #76, opened 2021-02-05
Fix commit message checks in push

mristin:
This patch tries to fix the workflow which checks commit messages in
pushes as currently it was not taken into account by GitHub.


mristin:
The checks are run on pushes from the main repository, but not on forks. The checks work as expected.

PR #81, opened 2021-02-11
Fix links and code blocks in the doc

mristin:
This patch makes various minor fixes to the documentation. Namely, the
external links with the named target were lacking the suffix underscore
("_"), and certain code blocks were missing double colon suffix 
(`..code-block:` instead of `.. code-block::`).


mristin:
@pschanely I manually inspected all the links. Now the issue with dangling links should be fixed. See the changelog page for how to specify the external links with named targets (I got that wrong initially).

pschanely:
Ah, I see. Great; thank you!

PR #79, opened 2021-02-10
Convert documentation to sphinx

mristin:
This patch converts the documentation from Markdown to Sphinx.

Additionally, the documentation is prepared to be published on
readthedocs.io.

PR #83, opened 2021-02-14
Fix doctesting for doc

mristin:
The `precommit.py` was checking on `*.md` documents in the `doc/`
directory, but the documentation was translated to RST + Sphinx in the
pull request #79.

### DIFF
--- a/doc/source/how_does_it_work.rst
+++ b/doc/source/how_does_it_work.rst
@@ -57,6 +57,8 @@ Let's Explore
 
 We can initialize a CrossHair object by giving it a name::
 
+    >>> from crosshair.libimpl.builtinslib import SmtInt
+
     >>> crosshair_x = SmtInt('x')
 
 We can access the ``.var`` attribute of any CrossHair object to get
@@ -70,6 +72,8 @@ the Z3 variable(s) that it holds:
 
 This takes the Z3 variable we just defined and adds one to it::
 
+    >>> import Z3
+
     >>> expr = crosshair_x.var + Z3.IntVal(1)
     >>> expr
     x + 1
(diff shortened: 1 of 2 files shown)


mristin:
@pschanely  could you please have a look at the failures? I only had a very quick look, but it seems to me that it is related to the recent implementation changes.

pschanely:
Oh dear. These examples aren't runnable for a few different reasons right now. I had imagined that rst testing would work more like a Jupyter notebook than a bunch of disjoint tests. I would also need to set up a few context managers around the whole thing to make it runnable.

Do you have any suggestions?

We could manually exclude the file for now. Though I think there are a total of zero real doctests in the documentation right now, so this phase isn't really doing anything at the moment.

mristin:
I think that the context is indeed reused, but I might be wrong. I'll try to make a simple test to verify that tonight.

pschanely:
Oh, possibly it does and I was just assuming that it doesn't!

Looks like we have [an extension](https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#skipping-tests-conditionally) already in our conf.py that would let me at least skip these for now. Though, I think those things only work if we run doctest through sphinx like `sphinx-build -b doctest ...`

mristin:
> Oh, possibly it does and I was just assuming that it doesn't!

I have just tested it manually. Doctest indeed works like a Jypiter notebook, the context is shared among the individual code snippets. I pushed another commit that imports the necessary structures, but the first snippet fails (hence `crosshair_x` is never properly initialized):

```
File "doc\source\how_does_it_work.rst", line 62, in how_does_it_work.rst
Failed example:
    crosshair_x = SmtInt('x')
Exception raised:
    Traceback (most recent call last):
      File "C:\users\rist\AppData\Local\Programs\Python\Python38\lib\doctest.py", line 1336, in __run
        exec(compile(example.source, filename, "single",
      File "<doctest how_does_it_work.rst[1]>", line 1, in <module>
        crosshair_x = SmtInt('x')
      File "C:\Users\rist\workspace\mristin\CrossHair\crosshair\libimpl\builtinslib.py", line 803, in __init__
        SmtIntable.__init__(self, smtvar, typ)
      File "C:\Users\rist\workspace\mristin\CrossHair\crosshair\libimpl\builtinslib.py", line 152, in __init__
        self.statespace = context_statespace()
      File "C:\Users\rist\workspace\mristin\CrossHair\crosshair\statespace.py", line 176, in context_statespace
        space = _THREAD_LOCALS.space
    AttributeError: '_thread._local' object has no attribute 'space'
```

I'm not familiar with the code base enough, but this seems like a bug to me?

Skipping pages for doctesting is actually even easier (we can manually exclude them in `precommit.py`). However, I would really recommend you to fix the doctests if possible, as tested documentation is really priceless.

pschanely:
Agreed. I need to add some context managers that do a few things before it'll fully work on its own. I think I can do what I want with some of the [testsetup](https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#directive-testsetup) stuff and explicitly call `__enter__`/`__exit__`. Do you want to just make this PR skip it, and I'll try to follow up later this week?

mristin:
I added a commit to skip the offending file (`how_does_it_work.rst`). Please squash and merge, and don't forget to remove the lines in `precommit.py` once the doctests are fixed ;-).

Issue #37, opened 2020-11-18
Add Support for Python 3.9


pschanely:
Python 3.9 makes some significant changes to type hinting, and CrossHair is likely unprepared. Needs investigation.

pschanely:
One blocker here is https://github.com/ilevkivskyi/typing_inspect/issues/65

Or, we decide to work around it.

pschanely:
Basic support for 3.9 has been added in 5267d2959df9904010ed9b3f2ab10d13e9f41c8d.

PR #89, opened 2021-02-27
Implement rfind, rsplit, replace, find

Rik-de-Kort:
Closing some items in #39.
Re-implements `find`.
Implements `rfind`, `rsplit`, and `replace`.


pschanely:
First, thanks so much for investing in this, and for being vocal about what's working and what isn't.

First, I suspect you're on python 3.9 and our battery of tests doesn't pass on that yet. (#37) Some of the work there is simple fixes like the mypy errors, but others (test_new_style_type_hints) are tests that I've added preemptively to ensure 3.9 features work properly with CrossHair.

All your mypy errors and most of your test failures I also get on 3.9, so if you're willing to setup again on 3.8, I suspect you'll get better results.

One problem I can't replicate locally with 3.9: the `Fraction` problem with `remove_outliers` (as part of `test_examples`, which just runs crosshair on everything in the examples directories); perhaps you'd be willing to dump the output of `crosshair check crosshair/examples/PEP316/correct_code/showcase.py --analysis_kind PEP316 -v` into a new issue for me?

If you ever see `... object has no attribute 'ctx'`, you can ignore that - it's something with z3 finalization that I don't think is terribly harmful.

Finally, if you think some failure isn't related to what you're doing, you can also just push to github and let it run the tests in github actions to see what happens there.

There is a fair amount of work I can do to improve test reproducibility (and runtime), so your feedback will help me figure out the highest priority things to look at there as well. Thanks again!

Rik-de-Kort:
No problem, happy to help :) I got MyPy working with 3.8. I still have 3 test failure but doesn't seem to be anything serious.

Can you provide some info on how to test the stuff I build? I see there is a list of patched string methods, so does my implementation of `SmtStr` *do* anything? How can I verify that I'm moving forward?

pschanely:
Great questions, and thanks for asking them - these are super helpful in terms of starting to assemble some real developer-facing documentation.

So when we define a method on SmtStr, we already know that `self` is symbolic. The arguments to that method may or may not be symbolic.

When we patch a builtin string method (like we do with `_str_join`), we know that `self` is concrete (i.e. a real string). The arguments might be symbolic though.

So sometimes, we'll define it in one place or the other or even both - those assumptions about what is and isn't symbolic can be pretty important in informing how we proceed.

In general, the highest priority is to handle the cases where data is most likely to be symbolic. In the case of split, usually the argument is concrete (often a literal like ":" or something). In the case of join, it's the opposite: usually `self` is a literal, and the argument to join is more likely to be symbolic. That's why we want to patch the builtin str for join, but for split, it's better to start with a method on SmtStr. (which is what you've done!)

But the point to drive home is that they aren't really fallbacks; one of the two approaches will get called depending on whether `self` is an `SmtStr` or a regular `str`.

Your understanding of the `check_ok` and `check_fail` test helpers is correct. There is also `check_unknown` for cases where we don't find a counterexample, but cannot search every path.

The general patten so far has been to exercise your methods by testing properties; like the tests you have right now. I have been thinking a bit about how to make some of these tests be a little more targeted, but don't have a good pattern for you to follow yet. If you comment out your new `rsplit` method, I suspect that your check_ok test would not pass, because CrossHair would just throw random strings at it, and cannot test every possible string. (check_unknown would pass though)

I would choose a postcondition for the `check_fail` that would be hard to hit randomly; for instance `post: _ != ["ab", "cd"]`: that way there's only one possible counterexample input, namely "ab:cd". CrossHair would not find this counterexample at HEAD, but hopefully will with your implementation!

Rik-de-Kort:
Thank you for the thorough explanation. I tried editing the `check_fail`, but that makes the tests run forever. I have pushed and here they have the same issue. I've included a stacktrace from a KeyboardInterrupt. Are we just looking at an exploding statespace that the solver can't handle?

<details>
<summary>Local stack trace</summary>
````Traceback (most recent call last):
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/_pytest/main.py", line 269, in wrap_session
    session.exitstatus = doit(config, session) or 0
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/_pytest/main.py", line 323, in _main
    config.hook.pytest_runtestloop(session=session)
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/pluggy/hooks.py", line 286, in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/pluggy/manager.py", line 93, in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/pluggy/manager.py", line 84, in <lambda>
    self._inner_hookexec = lambda hook, methods, kwargs: hook.multicall(
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/pluggy/callers.py", line 208, in _multicall
    return outcome.get_result()
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/pluggy/callers.py", line 80, in get_result
    raise ex[1].with_traceback(ex[2])
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/pluggy/callers.py", line 187, in _multicall
    res = hook_impl.function(*args)
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/_pytest/main.py", line 348, in pytest_runtestloop
    item.config.hook.pytest_runtest_protocol(item=item, nextitem=nextitem)
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/pluggy/hooks.py", line 286, in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/pluggy/manager.py", line 93, in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/pluggy/manager.py", line 84, in <lambda>
    self._inner_hookexec = lambda hook, methods, kwargs: hook.multicall(
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/pluggy/callers.py", line 208, in _multicall
    return outcome.get_result()
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/pluggy/callers.py", line 80, in get_result
    raise ex[1].with_traceback(ex[2])
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/pluggy/callers.py", line 187, in _multicall
    res = hook_impl.function(*args)
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/_pytest/runner.py", line 109, in pytest_runtest_protocol
    runtestprotocol(item, nextitem=nextitem)
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/_pytest/runner.py", line 126, in runtestprotocol
    reports.append(call_and_report(item, "call", log))
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/_pytest/runner.py", line 215, in call_and_report
    call = call_runtest_hook(item, when, **kwds)
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/_pytest/runner.py", line 254, in call_runtest_hook
    return CallInfo.from_call(
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/_pytest/runner.py", line 311, in from_call
    result: Optional[TResult] = func()
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/_pytest/runner.py", line 255, in <lambda>
    lambda: ihook(item=item, **kwds), when=when, reraise=reraise
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/pluggy/hooks.py", line 286, in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/pluggy/manager.py", line 93, in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/pluggy/manager.py", line 84, in <lambda>
    self._inner_hookexec = lambda hook, methods, kwargs: hook.multicall(
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/pluggy/callers.py", line 208, in _multicall
    return outcome.get_result()
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/pluggy/callers.py", line 80, in get_result
    raise ex[1].with_traceback(ex[2])
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/pluggy/callers.py", line 187, in _multicall
    res = hook_impl.function(*args)
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/_pytest/runner.py", line 162, in pytest_runtest_call
    item.runtest()
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/_pytest/unittest.py", line 321, in runtest
    self._testcase(result=self)  # type: ignore[arg-type]
  File "/usr/lib/python3.9/unittest/case.py", line 653, in __call__
    return self.run(*args, **kwds)
  File "/usr/lib/python3.9/unittest/case.py", line 593, in run
    self._callTestMethod(testMethod)
  File "/usr/lib/python3.9/unittest/case.py", line 550, in _callTestMethod
    method()
  File "/home/rik/CrossHair/crosshair/libimpl/builtinslib_test.py", line 489, in test_rsplit_fail
    self.assertEqual(*check_fail(f))
  File "/home/rik/CrossHair/crosshair/test_util.py", line 33, in check_fail
    states = [m.state for m in run_checkables(analyze_function(fn, options))]
  File "/home/rik/CrossHair/crosshair/core.py", line 739, in run_checkables
    collector.extend(checkable.analyze())
  File "/home/rik/CrossHair/crosshair/core.py", line 667, in analyze
    analysis = analyze_calltree(options, conditions)
  File "/home/rik/CrossHair/crosshair/core.py", line 961, in analyze_calltree
    call_analysis = attempt_call(
  File "/home/rik/CrossHair/crosshair/core.py", line 1329, in attempt_call
    detail = "false " + get_input_description(
  File "/home/rik/CrossHair/crosshair/core.py", line 1061, in get_input_description
    repr_str = repr(return_val)
  File "/home/rik/CrossHair/crosshair/util.py", line 291, in _eval_friendly_repr
    return _orig(obj)
  File "/home/rik/CrossHair/crosshair/core.py", line 121, in call_if_enabled
    return patched_fn(*a, **kw)
  File "/home/rik/CrossHair/crosshair/enforce.py", line 41, in wrapper
    return fn(*a, **kw)
  File "/home/rik/CrossHair/crosshair/core.py", line 868, in wrapper
    return original(*a, **kw)
  File "/home/rik/CrossHair/crosshair/libimpl/builtinslib.py", line 2314, in _repr
    return type(arg).__repr__(arg)
  File "/home/rik/CrossHair/crosshair/libimpl/builtinslib.py", line 1847, in __repr__
    return repr(self.__str__())
  File "/home/rik/CrossHair/crosshair/libimpl/builtinslib.py", line 1841, in __str__
    return self.statespace.find_model_value(self.var)
  File "/home/rik/CrossHair/crosshair/statespace.py", line 661, in find_model_value
    ModelValueNode(self._random, expr, self.solver)
  File "/home/rik/CrossHair/crosshair/statespace.py", line 516, in __init__
    if not solver_is_sat(solver):
  File "/home/rik/CrossHair/crosshair/statespace.py", line 291, in solver_is_sat
    debug("Unknown satisfiability. Solver state follows:\n", solver.sexpr())
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/z3/z3.py", line 6946, in sexpr
    return Z3_solver_to_string(self.ctx.ref(), self.solver)
  File "/home/rik/CrossHair/venv/lib/python3.9/site-packages/z3/z3core.py", line 3878, in Z3_solver_to_string
    r = _elems.f(a0, a1)
KeyboardInterrupt````
</details>

**Comment by pschanely** [comment]
Interesting! To be honest, I am not confident that LastIndexOf even works?
```
>>> z3.simplify(z3.LastIndexOf(z3.StringVal("asd:asd"), z3.StringVal(":")))
3
```
So far, so good. But:
```
>>> a=z3.String('a')
>>> z3.solve(z3.LastIndexOf(a, z3.StringVal(":")) == 3)
[a = ""]
# and, just to double check the empty case:
>>> z3.simplify(z3.LastIndexOf("", z3.StringVal(":")))
-1
```

**Comment by davidmrdavid** [comment]
Just dropping by since I was working on #42, which you referenced, to say not to worry about "stealing anyone's lunch" or anything like that 😄. My schedule got super hectic so I haven't had time to work on open source lately. Looking forward to seeing how you implement `rsplit` and others! ❤️

**Comment by pschanely** [comment]
@davidmrdavid no worries of course and happy to have you back anytime :)

I'll file a Z3 bug, but for the moment, it might be easiest to define rsplit in term of split (get all the pieces, and then re-join the left side if necessary). I think this will catch our check_fail case. (please re-fetch HEAD, as 66a44a0d10021e0b1e2d847a677274e62ddd1e9d is important for this case I think) Unfortunately, we'll need to change the check_ok test into a check_unknown, because there are an infinite number of possible splits working from the left side of the string. But not a huge deal: CrossHair is mostly just a bug-finding tool.

If you want to be even more ambitious, you can try implementing the rfind yourself by positing the existence of a new symbolic string that is a suffix of the given string, starts with the split string, and doesn't contain the split string again. An example of this is string multiply (e.g. `"foo" * 3`) where we [invent a new symbolic string](https://github.com/pschanely/CrossHair/blob/66a44a0d10021e0b1e2d847a677274e62ddd1e9d/crosshair/libimpl/builtinslib.py#L743) and use `space.add()` to inject as axioms into the solver that force it to match a regular expression (`foo*`) and have the length we expect. This is a little harder to get right, but I'm happy to help you along the way if you want to try it.

**Comment by Rik-de-Kort** [comment]
Hmm I wrote a vanilla z3 implementation which works. This example return 'sat, result=8'.
````
import z3
from crosshair.statespace import StateSpaceContext

space = z3.Solver()

with StateSpaceContext(space):
    def rfind(value, sub, start=None, end=None):
        start = 0 if start is None else start
        end = z3.Length(value) if end is None else end
        result = z3.Int('result')

        index_remaining = result + len(sub)
        last_match = z3.SubString(s, result, len(sub))
        remaining = z3.SubString(s, index_remaining, len(s) - index_remaining)
        found_match = z3.And(z3.Contains(last_match, sub), z3.Not(z3.Contains(remaining, sub)))
        no_match = z3.And(z3.Not(z3.Contains(value, sub)), result == -1)
        space.add(z3.Or(found_match, no_match))
        return result
    
    rfind('this is foobared!', 'foobar')
print(space.check(), space.model())
````

However, when I do the following with the current code for rfind, I get an incorrect result. It return 'sat, result=-1'.
````python
from crosshair.libimpl.builtinslib import SmtStr
from crosshair.statespace import StateSpaceContext

space = z3.Solver()
with StateSpaceContext(space):
    f = SmtStr('wow this is foobared')
    f.rfind('foobar')
    
print(space.check(), space.model())
````

Code for `rfind`:
````python
def rfind(self, substr, start=None, end=None):
        value = self.var
        sub = force_to_smt_sort(substr, SmtStr)
        start = 0 if start is None else force_to_smt_sort(start, SmtInt)
        end = z3.Length(value) if end is None else force_to_smt_sort(end, SmtInt)
        result = z3.Int("result")

        index_remaining = result + z3.Length(sub)
        last_match = z3.SubString(value, result, z3.Length(sub))
        remaining = z3.SubString(
            value, index_remaining, z3.Length(value) - index_remaining
        )
        found_match = z3.And(
            z3.Contains(last_match, sub), z3.Not(z3.Contains(remaining, sub))
        )
        no_match = z3.And(z3.Not(z3.Contains(value, sub)), result == -1)
        self.statespace.add(z3.Or(no_match, found_match))
        return result
````

**Comment by Rik-de-Kort** [comment]
Here's a snippet to test. Unfortunately, this currently spits out `result=-1`, rather than `result=8`.
````python
from crosshair.libimpl.builtinslib import SmtStr
from crosshair.statespace import StateSpaceContext

space = z3.Solver()
with StateSpaceContext(space):
    f = SmtStr('wow this is foobared')
    f.rfind('foobar')
    
space.check(), space.model()
````

Equivalent vanilla z3 code *does* work:
````python
space = z3.Solver()

with StateSpaceContext(space):
    def rfind(value, sub, start=None, end=None):
        start = 0 if start is None else start
        end = z3.Length(value) if end is None else end
        result = z3.Int('result')

        index_remaining = result + len(sub)
        last_match = z3.SubString(value, result, len(sub))
        remaining = z3.SubString(value, index_remaining, len(value) - index_remaining)
        found_match = z3.And(z3.Contains(last_match, sub), z3.Not(z3.Contains(remaining, sub)))
        no_match = z3.And(z3.Not(z3.Contains(value, sub)), result == -1)
        space.add(z3.Or(found_match, no_match))
        return result
    
    rfind('this is foobared!', 'foobar')
space.check(), space.model()
````

Any ideas?

**Comment by pschanely** [comment]
Wow, you've figured a lot out on your own - it's looking promising!

So, you are not the first one to hit this common gotcha about `SmtStr`: take a look at the resolved comment thread [here, just below line 1683](https://github.com/pschanely/CrossHair/pull/42/files/2ade81a69cf5dc87d72c5aaafe491bf710f1bc77#diff-765da293ca6f6070ffc34858feabc0ac841f5c82547ce356c669df131276b7b4R1683) (you have to expand the resolved conversation yourself - I can't directly link to it I suppose?) Anyway, TLDR is that you want to say `SmtStr(z3.StringVal("wow this is foobared"))`

**Comment by pschanely** [comment]
I'm also happy to leave comments on your in-progress code if you'd find that helpful.
(but I also don't want to inundate you with feedback if you don't want it yet)

**Comment by Rik-de-Kort** [comment]
Thanks for the compliment! I have a degree in mathematics so that helps ;)

Code appears to work with a `StringVal`, at least in my tests. I also fixed the return value in case `start` is present. 
I think now would be a good time to start commenting on the code itself. 
The tests still don't pass with strange messages (telling me input '' invalidates the post-condition which is patently false). I think it might have something to do with the fixed variable names? Should I turn those into SmtInt's?

File: crosshair/libimpl/builtinslib.py [hunk 1]
```diff
@@ -1916,6 +1916,31 @@ def find(self, substr, start=None, end=None):
...
+        last_match = z3.SubString(value, match_index, z3.Length(sub))
+        remaining = z3.SubString(
+            value, index_remaining, z3.Length(value) - index_remaining
+        )
+        found_match = z3.And(
+            z3.Contains(last_match, sub), z3.Not(z3.Contains(remaining, sub))
+        )
+        no_match = z3.And(z3.Not(z3.Contains(value, sub)), match_index == -1)
+        space.add(z3.Or(no_match, found_match))
+        return return_value
```
pschanely, line 1942:
```suggestion
        return SmtInt(return_value)
```
This wraps the z3 value into something that works more invisibly like a real Python integer. (the pattern you'll see in general is that we access a bunch of `.var` to get z3 values, do some stuff in z3, and then wrap everything back up in `Smt*` instances before returning.

crosshair/libimpl/builtinslib.py
```diff
@@ -1916,6 +1916,31 @@ def find(self, substr, start=None, end=None):
...
 
+    def rfind(self, substr, start=None, end=None):
+        space = self.statespace
+        sub = force_to_smt_sort(substr, SmtStr)
+
+        start = 0 if start is None else force_to_smt_sort(start, SmtInt)
+        end = z3.Length(value) if end is None else force_to_smt_sort(end, SmtInt)
+        value = z3.SubString(self.var, start, end - 1)
+        match_index = z3.Int("match_index")
+        return_value = z3.Int("return_value")
```
pschanely, line 1927:
```suggestion
        match_index = z3.Int("match_index" + space.uniq())
        return_value = z3.Int("return_value" + space.uniq())
```
Different values need to have different names in Z3. This just adds an underscore and some unique hex value to ensure it's unique. (otherwise you'll run into confusing behavior when calling rfind repeatedly!)

crosshair/libimpl/builtinslib.py
```diff
@@ -1916,6 +1916,31 @@ def find(self, substr, start=None, end=None):
...
+        index_remaining = match_index + z3.Length(sub)
+        last_match = z3.SubString(value, match_index, z3.Length(sub))
+        remaining = z3.SubString(
+            value, index_remaining, z3.Length(value) - index_remaining
+        )
+        found_match = z3.And(
+            z3.Contains(last_match, sub), z3.Not(z3.Contains(remaining, sub))
+        )
+        no_match = z3.And(z3.Not(z3.Contains(value, sub)), match_index == -1)
+        space.add(z3.Or(no_match, found_match))
```
pschanely, line 1941: A common tension in CrossHair: we can make things easier on the solver by adding branches in the path tree with `smt_fork()`. The downside is that we have to run more iterations. But especially for sting operations, Z3 gets to unsat pretty fast.

To make that idea concrete in your example, instead of this disjunction, we might say way up at the top:

```
if space.smt_fork(z3.Contains(self.var, sub)):
  ...
else:
  return -1
```
And then you can simplify some of the Z3 expressions because you can just assume that there is at least one match. CrossHair will try one branch on one iteration and the other on a future iteration.

To be clear, either approach works - this is really about tuning.

crosshair/libimpl/builtinslib.py
Rik-de-Kort: This was my hypothesis this morning when I saw the strange test results. I've made the change, and the test now passes! Added in latest commit. :)

crosshair/libimpl/builtinslib.py
Rik-de-Kort: Added in latest commit.

crosshair/libimpl/builtinslib.py
Rik-de-Kort: If we want to tune, we should have some measurements IMO. Good to know there are some options, though. I think they do make the code more readable.

Rik-de-Kort:
@pschanely tests are running fine on my computers. Anything up with the CI?

pschanely:
Hmm. `test_rsplit_fail` (using `z3.LastIndexOf`) hangs for me locally. But it passes for you?

Maybe we should try re-implementing `rsplit` using `rfind`?

Rik-de-Kort:
> Hmm. `test_rsplit_fail` (using `z3.LastIndexOf`) hangs for me locally. But it passes for you?
> 
> Maybe we should try re-implementing `rsplit` using `rfind`?

Ahhh my bad, I had only checked using `pytest -k rfind`. Will reimplement rsplit :)

Rik-de-Kort:
@pschanely been quiet for a while, but here's a thing. Note that I'm using the already tested `split`!

````python 
import z3
from crosshair.libimpl.builtinslib import SmtStr
from crosshair.statespace import SimpleStateSpace, StateSpaceContext

space = SimpleStateSpace()
with StateSpaceContext(space):
    s = SmtStr(z3.StringVal('split, this'))
    s.split(SmtStr(z3.StringVal(', ')))
````

 yields:
````python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-1-60879b23f928> in <module>
      6 with StateSpaceContext(space):
      7     s = SmtStr(z3.StringVal('split, this'))
----> 8     s.split(SmtStr(z3.StringVal(', ')))

~/CrossHair/crosshair/libimpl/builtinslib.py in split(self, sep, maxsplit)
   1975         if first_occurance == -1:
   1976             return [self]
-> 1977         ret = [self[: cast(int, first_occurance)]]
   1978         new_maxsplit = -1 if maxsplit == -1 else maxsplit - 1
   1979         ret.extend(self[first_occurance + 1 :].split(sep=sep, maxsplit=new_maxsplit))

~/CrossHair/crosshair/libimpl/builtinslib.py in __getitem__(self, i)
   1904 
   1905     def __getitem__(self, i):
-> 1906         idx_or_pair = process_slice_vs_symbolic_len(
   1907             self.statespace,
   1908             i,

~/CrossHair/crosshair/libimpl/builtinslib.py in process_slice_vs_symbolic_len(space, i, smt_len, fork_on_negative_index)
   1329             raise CrosshairUnsupported("slice steps not handled")  # TODO: handle this!
   1330         start = normalize_symbolic_index(start) if i.start is not None else z3.IntVal(0)
-> 1331         stop = normalize_symbolic_index(stop) if i.stop is not None else smt_len
   1332         return (start, stop)
   1333     else:

~/CrossHair/crosshair/libimpl/builtinslib.py in normalize_symbolic_index(idx)
   1306         elif fork_on_negative_index:
   1307             smt_idx = SmtInt._coerce_to_smt_sort(idx)
-> 1308             if idx >= 0:
   1309                 return smt_idx
   1310             else:

TypeError: '>=' not supported between instances of 'SmtInt' and int'
````
Not sure what's going awry here. I thought you might have some ideas.

**Comment by pschanely** [comment]
> Edit: got it, need to import `crosshair.core_and_libs`.

Damn, you're good.
Sometime, I'll need to ask you for a brain dump of the gotchas so that I can make some good dev docs. Also, the experience of experimenting at the command line could be improved a lot.

As you probably figured out, `cast` only is for static checking; at runtime it just returns its input. :)

> @pschanely been quiet for a while, but here's a thing. Note that I'm using the already tested `split`!

Very exciting! And, as always, happy for any time you're willing to invest.

**Comment by Rik-de-Kort** [comment]
> > Edit: got it, need to import `crosshair.core_and_libs`.
> 
> Damn, you're good.

Well the comment that said "it's easy to forget to import crosshair.core_and_libs" right above one of the offending lines helped. :P

> Sometime, I'll need to ask you for a brain dump of the gotchas so that I can make some good dev docs. Also, the experience of experimenting at the command line could be improved a lot.

I have some stuff in a jupyter notebook, this will probably grow over time. After I've gotten the hang of it, I might make the PR myself (but no promises).

**Comment by Rik-de-Kort** [comment]
There we go, working implementation of `rsplit`! :D

File: crosshair/libimpl/builtinslib.py [hunk 4]
```diff
@@ -1937,6 +1937,29 @@ def find(self, substr, start=None, end=None):
...
+        if space.smt_fork(z3.Contains(value, sub)):
+            match_index = z3.Int(f"match_index_{space.uniq()}")
+            last_match = z3.SubString(value, match_index, z3.Length(sub))
+            index_remaining = match_index + z3.Length(sub)
+            remaining = z3.SubString(
+                value, index_remaining, z3.Length(value) - index_remaining
+            )
+            space.add(
+                z3.And(
+                    z3.Contains(last_match, sub), z3.Not(z3.Contains(remaining, sub))
```
pschanely, line 1956: 1. RE `z3.Contains(last_match, sub)`: Even more than containment? Are `last_match` and `sub` guaranteed to be the same string at this point?
2. RE `3.Not(z3.Contains(remaining, sub))`: help me think about whether matches to the right could overlap with the matched portion of the string. e.g. do we prevent `"aaa".rfind("aa")` from returning zero?

crosshair/libimpl/builtinslib.py
```diff
@@ -1967,6 +1990,24 @@ def split(self, sep: Optional[str] = None, maxsplit: int = -1):
         ret.extend(self[first_occurance + 1 :].split(sep=sep, maxsplit=new_maxsplit))
         return ret
 
+    def rsplit(self, sep: Optional[str] = None, maxsplit: int = -2):
+        if sep is None:
+            return self.__str__().split(sep=sep, maxsplit=maxsplit)
```
pschanely, line 1995:
```suggestion
            return self.__str__().rsplit(sep=sep, maxsplit=maxsplit)
```

crosshair/libimpl/builtinslib.py
```diff
@@ -1967,6 +1990,24 @@ def split(self, sep: Optional[str] = None, maxsplit: int = -1):
         ret.extend(self[first_occurance + 1 :].split(sep=sep, maxsplit=new_maxsplit))
         return ret
 
+    def rsplit(self, sep: Optional[str] = None, maxsplit: int = -2):
```
pschanely, line 1993: Is the default of maxsplit supposed to be -1?

crosshair/libimpl/builtinslib.py
```diff
@@ -1967,6 +1990,24 @@ def split(self, sep: Optional[str] = None, maxsplit: int = -1):
...
+            return self.__str__().split(sep=sep, maxsplit=maxsplit)
+        smt_sep = force_to_smt_sort(sep, SmtStr)
+        if not isinstance(maxsplit, Integral):
+            raise TypeError
+        if maxsplit == 0:
+            return [self]
+        last_occurence = self.rfind(sep)
+        if last_occurence == -1:
+            return [self]
+        new_maxsplit = -1 if maxsplit == -1 else maxsplit - 1
```
pschanely, line 2004:
```suggestion
        new_maxsplit = -1 if maxsplit < 0 else maxsplit - 1
```

I think this is wrong in my version of `split()` too - Python seems to allow the user to pass in `maxsplit=-2` and it should be treated the same as `-1`. But you don't need to bother fixing `split()`.

crosshair/libimpl/builtinslib.py
```diff
@@ -1967,6 +1990,24 @@ def split(self, sep: Optional[str] = None, maxsplit: int = -1):
...
+        remaining = self[: cast(int, last_occurence)]
+        ret = self[:last_occurence].rsplit(sep, new_maxsplit)
+        index_after = SmtInt(z3.Length(smt_sep)) + last_occurence
```
pschanely, line 2007:
```suggestion
        index_after = len(sep) + last_occurence
```
Equivalent (I believe). Just a tad simpler. (and it doesn't use z3 in case `sep` is a concrete string)

pschanely (commented):
This is very exciting - thank you!
And, apologies in advance for the nits here; I think there are potentially a lot of edge cases.

I also think it's ok if we merge without assurance that it's perfect. I am going to work on a better fuzz test setup that might help us uncover them before the next release.

crosshair/libimpl/builtinslib.py
```diff
@@ -1937,6 +1937,29 @@ def find(self, substr, start=None, end=None):
         start = 0 if start is None else force_to_smt_sort(start, SmtInt)
         return SmtInt(z3.IndexOf(smt_mystr, smt_substr, start))
 
+    def rfind(self, substr, start=None, end=None):
+        space = self.statespace
+        sub = force_to_smt_sort(substr, SmtStr)
+
+        start = 0 if start is None else force_to_smt_sort(start, SmtInt)
+        end = z3.Length(self.var) if end is None else force_to_smt_sort(end, SmtInt)
+        value = z3.SubString(self.var, start, end - 1)
```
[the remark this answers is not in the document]
Rik-de-Kort: Sounds like it. Will include a bounds check. Though I think `find` also has this problem.

crosshair/libimpl/builtinslib.py
Rik-de-Kort: 1. Yes, because `last_match` is defined by taking a substring of length `sub`, so it suffices to check that `last_match` contains `sub`. I really should take a good look at the code to get it in a bit more readable order.
2. Good catch! Redefining `index_remaining = match_index+1` will take care of this problem in general, I think.

crosshair/libimpl/builtinslib.py
Rik-de-Kort: Yeah I ended up at this construction by fighting with the error messages for half an hour. Still need more solid intuition for what happens in z3-land and what happens in Python-land and how `Smt` brokers between those.

crosshair/libimpl/builtinslib.py
Rik-de-Kort: Oops

Rik-de-Kort:
Thanks for the review and your continuing kind words! In regards to merging it even if it isn't perfect: I think it's good practice to figure out the details of the system to get all these funny edge cases working. I think I will do some more work on `rfind` tomorrow.

Rik-de-Kort:
My goodness, that bounds checking got pretty out of hand. Ended up doing a sort of pattern match on `start` and `end` rather than going at it separately (because then the types don't line up). Is there a way we can make it simpler?

crosshair/libimpl/builtinslib.py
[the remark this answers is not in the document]
pschanely: Ah yeah, three thoughts.
(1) Looks to me like Python essentially behaves as though start and end are clamped to the string bounds; and I think we can do the same here; we just have to be careful that whatever we pass to z3.SubString is clamped appropriately.
(2) It might be a little easier to do this preprocessing in regular python, before getting z3 values out; maybe something like this:
```
mylen = len(self)
if start > mylen or end < 0:
  return -1
if start < 0:
  start = 0
if end > mylen:
  end = mylen
start = force_to_smt_sort(start, SmtInt)
end = force_to_smt_sort(end, SmtInt)
value = z3.SubString(self.var, start, end - start)
```
(3) something I hadn't noticed before; the last argument to z3.SubString is a length, and an ending index, so I think we want `end - start` there.

crosshair/libimpl/builtinslib.py
pschanely: This is tricky - there are three different kinds of values you might have for any python type. Check out [the table here](https://crosshair.readthedocs.io/en/latest/how_does_it_work.html#the-basic-idea) if you haven't already.

It's worth noting that the Smt* classes emulate native python values, so if you know you have those, you can just use regular python on them. 

There are a few different ways to get a z3 value. If you know you've got an Smt* instance, you can just access the `.var` attribute of it. The `coerce_to_smt_sort(...)` / `force_to_smt_sort(...)` functions are helpers that can take either Smt* instances or native Python values and give you back the appropriate z3 value.

When going the other way, you just construct the Smt*() instance and pass in the z3 value.

Not sure whether any of this helps or even makes any sense!

crosshair/libimpl/builtinslib.py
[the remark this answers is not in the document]
Rik-de-Kort: (1) Negative arguments to `find` are supported. `'foo'.find('o', -1)` will return `2`. [According to the documentation](https://docs.python.org/3/library/stdtypes.html#str.find), the semantics are `s.find(sub, start, end) = s[start:end].find(sub) + start`.  
(2) Alright! With this it's also a lot easier to implement the behaviour above.  
(3) and (4) Already have an offset in there ;) I'll also have a look at `find`, see if it needs anything.

crosshair/libimpl/builtinslib.py
Rik-de-Kort: Ah so where my thinking was going funky was that I assumed `coerce_to_smt_sort(...)` would give me an Smt-object, not a z3-object!

crosshair/libimpl/builtinslib.py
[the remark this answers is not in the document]
pschanely: Ah! I hadn't noticed that it was supposed to act like a regular slice. Good idea! Yes - just slice the string normally.

This is common in CrossHair - it's almost always better to write things in terms of stuff that already works with symbolics.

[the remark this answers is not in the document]
Rik-de-Kort: Some odd boundary conditions still, like `z3.SubStr` is ill-defined with negative length argument, but overall code is looking much cleaner now. I also took the liberty of adjusting `find` because it was returning the wrong thing on negative arguments.

crosshair/libimpl/builtinslib.py
Rik-de-Kort: I think it's a good idea to fix `split` while I'm in the neighbourhood. :)

crosshair/libimpl/builtinslib.py
```diff
@@ -1929,13 +1929,39 @@ def __getitem__(self, i):
...
+        sub = force_to_smt_sort(substr, SmtStr)
+        if self.statespace.smt_fork(z3.Contains(value, sub)):
+            return SmtInt(z3.IndexOf(value, sub, 0) + start)
+        else:
+            return -1
+
+    def rfind(self, substr, start=None, end=None):
+        start = 0 if start is None else start + len(self) if start < 0 else start
+        if end is not None and end <= start:
+            return -1
```
pschanely, line 1945: At first I thought that the slice would take care of this for us?

Then I noticed this, which made me think some sort of special handling might be necessary. Sigh!:
```
>>> 'abcd'[2:1].find('')
0
>>> 'abcd'.find('', 2, 1)
-1
```

```

Comment (Rik-de-Kort) [reply crosshair/libimpl/builtinslib.py:1945]: Yup, noticed the same thing, which is why that line is in there. I think we've reached the "right" implementation. However, the fuzz test still finds issues. How do I find out what inputs are causing it?

Comment (pschanely) [reply crosshair/libimpl/builtinslib.py:1945]: I want to refactor that to use pytest.parametrize, but for now, this is what I do:
```
python crosshair/fuzz_core_test.py FuzzTest.test_[whatever] -v 2>&1 >log.txt
```
And search log.txt for several asterisks to find to the failing example.

Note that when exceptions are thrown, we try to be consistent with the exception type, but the exception message doesn't have to match.

Comment (Rik-de-Kort) [reply crosshair/libimpl/builtinslib.py:1948]: It's finding some obviously pathological input which complicates things a bit, but that's ducktyping for ya. Funnily enough it also didn't trigger when making the slice object, but only when trying to use it.

**Comment by Rik-de-Kort** [comment]
All checks passed, hurrah!

**Comment by pschanely** [comment]
So great. Looks like you've made good progress on replace() so let's add a test and then we'll merge this chunk?

Also, feel free to add yourself to `contributors.rst` and link it to whatever you like!

File: doc/source/contributors.rst [hunk 11]
```diff
@@ -6,3 +6,4 @@ Contributors
 * `Edward Haigh <https://github.com/oneEdoubleD>`_
 * `Saul Shanabrook <https://github.com/saulshanabrook/>`_
 * `Marko Ristin <https://github.com/mristin>`_
+* `Rik de Kort <https://github.com/Rik-de-Kort>`
```
pschanely, line 9:
```suggestion
* `Rik de Kort <https://github.com/Rik-de-Kort>`_
```
ReStructuredText is annoying - we need a trailing underscore here.

PR #93, opened 2021-03-07
Update docs for VS Code extension.

pschanely:
Add to integrations page, and list it in the latest news.
Remove some docs content that was emacs-specific.


mristin (approved):
LGTM!

PR #95, opened 2021-03-18
Implement simple string methods

Rik-de-Kort:
This PR continues work on #39

### DIFF
--- a/crosshair/libimpl/builtinslib.py
+++ b/crosshair/libimpl/builtinslib.py
@@ -1990,6 +1990,13 @@ class SymbolicStr(AtomicSymbolicValue, SymbolicSequence, AbcString):
         else:
             return -1
 
+    def rindex(self, substr, start=None, end=None):
+        result = self.rfind(substr, start, end)
+        if result == -1:
+            raise ValueError
+        else:
+            return result
+
     def replace(self, old, new, count=-1):
         if not isinstance(old, str) or not isinstance(new, str):
             raise TypeError
--- a/crosshair/libimpl/builtinslib.py
+++ b/crosshair/libimpl/builtinslib.py
@@ -2093,9 +2093,8 @@ class SymbolicStr(AtomicSymbolicValue, SymbolicSequence, AbcString):
 
     def zfill(self, width):
         if self.startswith("+") or self.startswith("-"):
-            return self[0] + self[1:].zfill(width-1)
+            return self[0] + self[1:].zfill(width - 1)
         return "0" * max(0, width - len(self)) + self
-            
 
 
 _CACHED_TYPE_ENUMS: Dict[FrozenSet[type], z3.SortRef] = {}
--- a/crosshair/libimpl/builtinslib.py
+++ b/crosshair/libimpl/builtinslib.py
@@ -2041,7 +2041,9 @@ class SymbolicStr(AtomicSymbolicValue, SymbolicSequence, AbcString):
             return [self]
         ret = [self[: cast(int, first_occurance)]]
         new_maxsplit = -1 if maxsplit < 0 else maxsplit - 1
-        ret.extend(self[first_occurance + 1 :].split(sep=sep, maxsplit=new_maxsplit))
+        ret.extend(
+            self[first_occurance + len(sep) :].split(sep=sep, maxsplit=new_maxsplit)
+        )
         return ret
 
     def rsplit(self, sep: Optional[str] = None, maxsplit: int = -1):
@@ -2061,6 +2063,13 @@ class SymbolicStr(AtomicSymbolicValue, SymbolicSequence, AbcString):
         index_after = len(sep) + last_occurence
         ret.append(self[index_after:])
         return ret
+
+    def partition(self, sep: str):
+        result = self.split(sep, maxsplit=1)
+        if len(result) == 1:
+            return (self, "", "")
+        elif len(result) == 2:
+            return (result[0], sep, result[1])
 
 
 _CACHED_TYPE_ENUMS: Dict[FrozenSet[type], z3.SortRef] = {}
--- a/crosshair/libimpl/builtinslib.py
+++ b/crosshair/libimpl/builtinslib.py
@@ -1953,6 +1953,9 @@ class SymbolicStr(AtomicSymbolicValue, SymbolicSequence, AbcString):
             smt_result = z3.Extract(self.var, idx_or_pair, 1)
         return SymbolicStr(smt_result)
 
+    def count(self, substr, start=None, end=None):
+        return len(self[start:end].split(substr)) - 1
+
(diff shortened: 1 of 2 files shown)


Rik-de-Kort:
Here's an example that produces the wrong result. It appears to work find with ASCII. Do we have to hold off on implementing this until we can get full unicode support in `SmtStr`/z3?
````python
import z3
from crosshair.libimpl.builtinslib import SmtStr
from crosshair.statespace import SimpleStateSpace, StateSpaceContext
import crosshair.core_and_libs 

space = SimpleStateSpace()
with StateSpaceContext(space):
    s = SmtStr(z3.StringVal('Â'))
    print(s.join(['a', 'b', 'c']), 'Â'.join(['a', 'b', 'c']))
````
Output
````python
a\u{ffffffc2}b\u{ffffffc2}c aÂbÂc
````

**Comment by pschanely** [comment]
My experience is that at least our version of z3 has issues with unicode, and have been working under the idea that (for the moment) SmtStr is only for ASCII. (well, technically a little larger - the codepoints 0-255)

You may have noticed that I'm using sequences of bitvectors in places to try and keep z3 from attempting to use other values.

In spite of this restriction, I think we should always consider it a bug if Crosshair produces a counterexample that doesn't reproduce.

**Comment by pschanely** [comment]
A heads-up (and apology). I'm trying to make the naming less confusing: 8f87df3a8d14e3a681dd6ced70b5416839a983b9 performs a large rename of `Smt*` to `Symbolic*`. So you'll have to rebase and resolve some conflicts - sorry!

**Comment by Rik-de-Kort** [comment]
> A heads-up (and apology). I'm trying to make the naming less confusing: [8f87df3](https://github.com/pschanely/CrossHair/commit/8f87df3a8d14e3a681dd6ced70b5416839a983b9) performs a large rename of `Smt*` to `Symbolic*`. So you'll have to rebase and resolve some conflicts - sorry!

No problem, I think it's a great idea!

**Comment by Rik-de-Kort** [comment]
> My experience is that at least our version of z3 has issues with unicode, and have been working under the idea that (for the moment) SmtStr is only for ASCII. (well, technically a little larger - the codepoints 0-255)
> 
> You may have noticed that I'm using sequences of bitvectors in places to try and keep z3 from attempting to use other values.

Based on a very brief excursion into the land of z3 and unicode, the issue appears whenever we try to materialize a string value (for instance, by printing it). I think for now not something I want to spend an inordinate amount of time on.

I'll continue on with the string methods and ping you when done.

**Comment by Rik-de-Kort** [comment]
@pschanely any idea why CrossHair can find the counterexample on 3.8 but not on 3.7?

**Comment by pschanely** [comment]
OK - apologies for the delays here!
I think the `test_partition_ok` and `test_rpartition_ok` issues are coming from a bug in the patching mechanism; they pass for me with a fix I just pushed in c7b658cea619ed0cbbef1f65f53402d60d2d1917.

`test_rpartition_fail` is more strange - I see an unknown satisfiability result from Z3 in the log, which happens where I'd expect to find the counterexample. I am less sure what to do here, but I'll continue to poke at it.

**Comment by pschanely** [comment]
Ah, silly me, `test_rpartition_fail` just needs a little more time. (string reasoning is expensive I suppose!) For now, you can do something like this:
```
self.assertEqual(*check_fail(f, AnalysisOptionSet(per_path_timeout=5)))
```
`per_path_timeout` controls the max seconds we'll spend on a single execution before giving up.
`per_condition_timeout` controls the max seconds we'll spend over all executions for checking one postcondition.

**Comment by Rik-de-Kort** [comment]
Thanks for the help! I don't think it's a good idea to add a flaky test and then rely on the timeout for it to pass (because that can still be ruined by poor computation speed), so I've taken it out for now. Maybe in some next PR we'll do some performance benchmarking, eh?

I think I'd like to call this ready for merging. :)

File: crosshair/libimpl/builtinslib.py [hunk 1]
```diff
@@ -1990,71 +2034,72 @@ def rfind(self, substr, start=None, end=None):
...
 
+    def startswith(self, substr):
+        smt_substr = force_to_smt_sort(substr, SymbolicStr)
+        return SymbolicBool(z3.PrefixOf(smt_substr, self.var))
+
+    def zfill(self, width):
+        if not isinstance(width, int):
+            raise TypeError
+        if self.startswith("+") or self.startswith("-"):
+            return self[0] + self[1:].zfill(width - 1)
```
pschanely, line 2100: Repeated lead characters might be a gotcha here!: 
```
>>> '++12'.zfill(10)
'+000000+12'
```

crosshair/libimpl/builtinslib_test.py
```diff
@@ -441,6 +441,13 @@ def f(a: str) -> int:
 
         self.assertEqual(*check_ok(f))
 
+    def test_ljust_fail(self) -> None:
+        def f(s: str) -> str:
+            """ post: _ == len(s) """
+            return s.ljust(3, "x")
+
+        self.assertEqual(*check_fail(f))
```
pschanely, line 449: So, this fails, but not in the way you intended, I suspect. `ljust` will return a string, so that won't ever equal the length of `s`.
Same for `rjust` below.

crosshair/libimpl/builtinslib_test.py
```diff
@@ -455,6 +462,30 @@ def f(a: str) -> int:
...
+            try:
+                return a.rindex("abc")
+            except ValueError:
+                return 0
+
+        self.assertEqual(*check_fail(f))
+
+    def test_rindex_err(self) -> None:
+        def f(a: str) -> int:
+            """ post: _ == -1 """
```
pschanely, line 477: Consider `post: True` to be extra sure that we don't accidentally find a counterexample before we hit the exception.

pschanely (commented):
Amazing. This is a lot of stuff! A few minor comments.

Sounds good on the testing. Later this week, I'd like to forward you a potential new test pattern for how we might expand our coverage, reduce flakes, and improve test time. I'd be curious about your opinions on it!

crosshair/libimpl/builtinslib.py
Rik-de-Kort: Refactored to explicitly use only one lead character. Python semantics are weird sometimes.

crosshair/libimpl/builtinslib_test.py
Rik-de-Kort: Oh yeah definitely not what I was going for, good catch!

crosshair/libimpl/builtinslib_test.py
Rik-de-Kort: Done in latest commit.

Rik-de-Kort:
Thanks for the review. Definitely up for having a look at the test pattern. :)

PR #98, opened 2021-03-26
Add self-tests for string methods

pschanely:
@Rik-de-Kort Spent a little time thinking about testing recently, and took some [notes here](https://github.com/pschanely/CrossHair/discussions/97).

I've been trying to figure out how to efficiently test all these tricky edge cases, and I think the approach in `builtinslib_ch_test.py` is reasonably effective (though I suspect still insufficient). I've left a variety of cases commented that aren't quite passing yet.

I also did some refactoring of `rfind` here: partly to cover some more of those edge cases, but also to try and get around some of the z3 unknown results we've been seeing. I have a suspicion that the string solver doesn't love to reason about numeric indices: it's possible that doing so requires the linear arithmetic solver to cooperate with the string solver, and things get harder at that point.

Anyway, I'd be curious about your opinions on, well, just about any of this.

### DIFF
--- a/additional-verbs-in-imperative-mood.txt
+++ b/additional-verbs-in-imperative-mood.txt
@@ -37,3 +37,4 @@ fork
 isolate
 redesign
 revert
+disable
--- a/crosshair/core.py
+++ b/crosshair/core.py
@@ -188,7 +188,7 @@ class _StandaloneStatespace(ExitStack):
         self.enter_context(Patched())
         self.enter_context(StateSpaceContext(space))
         self.enter_context(COMPOSITE_TRACER)
-        return self
+        return space
 
 
 standalone_statespace = _StandaloneStatespace()
--- a/crosshair/libimpl/builtinslib.py
+++ b/crosshair/libimpl/builtinslib.py
@@ -1330,20 +1330,19 @@ class SymbolicFrozenSet(SymbolicSet):
 
 
 def process_slice_vs_symbolic_len(
-    space: StateSpace, i: slice, smt_len: z3.ExprRef, fork_on_negative_index=False
+    space: StateSpace,
+    i: Union[int, slice],
+    smt_len: z3.ExprRef,
 ) -> Union[z3.ExprRef, Tuple[z3.ExprRef, z3.ExprRef]]:
     def normalize_symbolic_index(idx) -> z3.ExprRef:
         if type(idx) is int:
             return z3.IntVal(idx) if idx >= 0 else (smt_len + z3.IntVal(idx))
-        elif fork_on_negative_index:
+        else:
             smt_idx = SymbolicInt._coerce_to_smt_sort(idx)
             if idx >= 0:
                 return smt_idx
             else:
                 return smt_len + smt_idx
-        else:
-            smt_idx = SymbolicInt._coerce_to_smt_sort(idx)
-            return z3.If(smt_idx >= z3.IntVal(0), smt_idx, smt_len + smt_idx)
 
     if isinstance(i, (int, SymbolicInt)):
         smt_i = SymbolicInt._coerce_to_smt_sort(i)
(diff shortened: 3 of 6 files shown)


Rik-de-Kort:
This will be a bit stream-of-consciousness-like, so feel free to ask any questions as we go along.

The thing about `test_rfind_fail` and friends is that they are much more "integration-like" than "unit-like". They answer the question "Can CrossHair correctly find failing/passing properties?" rather than "Does the symbolic implementation correspond to the realized implementation?": they test things about the end-to-end pipeline that is CrossHair. This is essentially the first point in your notes.  
I'm happy to see we can verify symbolic execution more directly: the branching issue is not all that relevant probably since we know what's going on with unit tests anyway.   
One idea: what about using Hypothesis to augment `builtinslib_ch_test`? Hypothesis is a lot more mature and may be able to find bugs that we still have trouble finding. 

I think the core of it is:
- Is z3 the bottleneck, or is there something in the rest of our code that's taking a lot of time? How do we measure this, what is the benchmark?
- How do we improve z3-solver performance? I suppose we "program to" the z3 solver, using clever tricks to make it more efficient or narrow the state space. Additionally, we could view programming z3 directly as a sort of "assembly", adding in direct logic programming over writing some Python code if we see performance is lacking.

I just ran pytest-profile (forgot to pull, so results are on a bit older branch).
On a higher level: `fuzz_core_test` takes a lot of time, like 30 percent. The other monster is `check_examples`, clocking in at 13 percent which I'm not sure what it does.  This needs some disentangling by module IMO: the 13 percent is not apples to apples since that's running a lot of tests at once.  
On the low level: it seems the test suite spends about 25 percent in z3 actually solving things, and 20 percent on calls to `__bool__`, half of which are coming from some `__init__`. There's also a fair bit of `eval` in there but I'm unsure how to interpret that.

I think the profile is hampered by timeouts, because there's a cutoff for tests that "run too long". However, things that run too long might run quadratically or exponentially, and the cutoff prevents us from finding that out. Just something to consider when interpreting the results.

Here's the pytest-profile graph:
combined

pschanely:
Super fun. Some reactions below.
> One idea: what about using Hypothesis to augment builtinslib_ch_test? Hypothesis is a lot more mature and may be able to find bugs that we still have trouble finding.

Yes. I think hypothesis is a great candidate for replacing that `fuzz_core_test` in particular, since that one is trying to start with concrete values, and attempts to confirm that an equivalent symbolic behaves the same.

The pytest profile is very cool; it may not work against current HEAD, as we're using `sys.settrace` to do a variety of things inside CrossHair now. Some thoughts:
* Work under `__bool__` is pretty expected - these are called on SymbolicBool instances when Python forces us to make a decision, usually just before a branch. Generally Z3 will be invoked inside these.
* `check_examples` runs CrossHair on each of the files under the `examples/` directory, ensuring we get the expected results. It's a mix of an integration test and documentation test, I suppose. I have a few ideas about how to improve these tests a bit but probably won't try to type those thoughts out immediately.
* IIRC, one of the biggest problems may be that we don't abort immediately when finding a counterexample. Because in real use cases counterexamples are rare, and if we can find multiple counterexamples, we can prefer to display the most succinct one. Maybe I should revert that decision, or do something special in testing scenarios. Some piece of logic explicitly for counterexample minimization would do a much better job than what I've got currently.
* I am fairly optimistic about our ability to optimize CrossHair - there is likely a fair amount of headroom available. But at this stage I suspect our biggest QA headroom is in using test time effectively, rather than making CrossHair itself more efficient - I haven't been thoughtful about using test time effectively until very recently.

To that end, I firmly agree on your point about integration testing: For new tests, I'm going to try out this `standalone_statespace()` + `is_possible()` idea and see how that goes.

Anyway, probably will commit this PR, but I'd love to continue this discussion in #97.

Issue #94, opened 2021-03-11
Constructor compatibility for CrossHair symbolics


pschanely:
**Expected vs actual behavior**

CrossHair symbolic types cannot be constructed like their native counterparts are. (they take args like the z3 identifier for a symbolic, or a z3 expression). Mostly that's fine, but causes trouble when the code under analysis assumes it can take the type of a value, and construct a new value with the same type; e.g. `type(listarg)([1,2,3])`

This will create spurious errors, like in this [crosshair playground example](https://crosshair-web.org/?crosshair=0.1&python=3.8&gist=fe8098cb7c938bed3aa6d5008600837a&flags=verbose)

There are a couple of broad approaches for solving this problem, but I'm not ready to pull the trigger yet on those. One promising avenue is to patch `type()`, but this requires us to be very strict about what parts of the codebase should and shouldn't be patched - it's a bit of work.

Rik-de-Kort:
Bit boneheaded maybe, but how about adding "vanilla" constructor capabilities to `SmtInt` and friends? Like evaluating `SmtInt(5)` gives you the same thing as `SmtInt(z3.IntVal(5))`.  
It would be a bit better than patching `type`, since (I think) we rely on the type machinery in various parts of the codebase.  
It'd also remove the stumbling block that both I and another dev game across.

pschanely:
Not boneheaded at all - that's a viable approach. One gotcha is `SmtStr()` where a plain string argument means that it's creating a new symbolic with a z3 value having that name. Potentially, to work around this, we could migrate internal constructor usages to use special keyword arguments when creating using a z3 name. Happy to give you feedback on a more concrete proposal if you want to explore this.


Some background context about patching follows (which has been HEAVILY refactored with commits that I've pushed in the last few minutes!). This is just to give you an idea about how we deal with the fact that when running user code, we want symbolics to be invisible, and when we are inside CrossHair, we do not.

[...]

PR #99, opened 2021-04-01
Add constructor compatibility to symbolics

Rik-de-Kort:
Working on #94

### DIFF
--- a/crosshair/libimpl/builtinslib.py
+++ b/crosshair/libimpl/builtinslib.py
@@ -503,13 +503,13 @@ def setup_binops():
     # Note that we don't want this when `other` is a boolean, but that
     # case will be overridden in the booleans section below.
     def _(a: SymbolicBool, b: Number):
-        return (SymbolicInt(z3.If(a.var, 1, 0)), b)
+        return (SymbolicInt.from_z3(z3.If(a.var, 1, 0)), b)
 
     setup_promotion(_, _ALL_OPS)
 
     # Implicitly upconvert symbolic ints to floats.
     def _(a: SymbolicInt, b: Union[float, FiniteFloat, SymbolicFloat, complex]):
-        return (SymbolicFloat(z3.ToReal(a.var)), b)
+        return (SymbolicFloat.from_z3(z3.ToReal(a.var)), b)
 
     setup_promotion(_, _ARITHMETIC_AND_COMPARISON_OPS)
 
@@ -533,53 +533,53 @@ def setup_binops():
 
     # float
     def _(op: BinFn, a: SymbolicFloat, b: SymbolicFloat):
-        return SymbolicFloat(apply_smt(op, a.var, b.var))
+        return SymbolicFloat.from_z3(apply_smt(op, a.var, b.var))
 
     setup_binop(_, _ARITHMETIC_OPS)
 
     def _(op: BinFn, a: SymbolicFloat, b: SymbolicFloat):
-        return SymbolicBool(apply_smt(op, a.var, b.var))
+        return SymbolicBool.from_z3(apply_smt(op, a.var, b.var))
 
     setup_binop(_, _COMPARISON_OPS)
 
     def _(op: BinFn, a: SymbolicFloat, b: FiniteFloat):
-        return SymbolicFloat(apply_smt(op, a.var, z3.RealVal(b)))
+        return SymbolicFloat.from_z3(apply_smt(op, a.var, z3.RealVal(b)))
 
     setup_binop(_, _ARITHMETIC_OPS)
 
     def _(op: BinFn, a: FiniteFloat, b: SymbolicFloat):
-        return SymbolicFloat(apply_smt(op, z3.RealVal(a), b.var))
+        return SymbolicFloat.from_z3(apply_smt(op, z3.RealVal(a), b.var))
 
     setup_binop(_, _ARITHMETIC_OPS)
 
     def _(op: BinFn, a: SymbolicFloat, b: FiniteFloat):
-        return SymbolicBool(apply_smt(op, a.var, z3.RealVal(b)))
+        return SymbolicBool.from_z3(apply_smt(op, a.var, z3.RealVal(b)))
 
     setup_binop(_, _COMPARISON_OPS)
 
     # int
     def _(op: BinFn, a: SymbolicInt, b: SymbolicInt):
-        return SymbolicInt(apply_smt(op, a.var, b.var))
+        return SymbolicInt.from_z3(apply_smt(op, a.var, b.var))
 
     setup_binop(_, _ARITHMETIC_AND_BITWISE_OPS)
 
     def _(op: BinFn, a: SymbolicInt, b: SymbolicInt):
-        return SymbolicBool(apply_smt(op, a.var, b.var))
+        return SymbolicBool.from_z3(apply_smt(op, a.var, b.var))
 
     setup_binop(_, _COMPARISON_OPS)
 
     def _(op: BinFn, a: SymbolicInt, b: int):
-        return SymbolicInt(apply_smt(op, a.var, z3.IntVal(b)))
+        return SymbolicInt.from_z3(apply_smt(op, a.var, z3.IntVal(b)))
 
     setup_binop(_, _ARITHMETIC_AND_BITWISE_OPS)
 
     def _(op: BinFn, a: int, b: SymbolicInt):
-        return SymbolicInt(apply_smt(op, z3.IntVal(a), b.var))
+        return SymbolicInt.from_z3(apply_smt(op, z3.IntVal(a), b.var))
 
     setup_binop(_, _ARITHMETIC_AND_BITWISE_OPS)
 
     def _(op: BinFn, a: SymbolicInt, b: int):
-        return SymbolicBool(apply_smt(op, a.var, z3.IntVal(b)))
+        return SymbolicBool.from_z3(apply_smt(op, a.var, z3.IntVal(b)))
 
     setup_binop(_, _COMPARISON_OPS)
 
@@ -604,17 +604,17 @@ def setup_binops():
 
     # bool
     def _(op: BinFn, a: SymbolicBool, b: SymbolicBool):
-        return SymbolicBool(apply_smt(op, a.var, b.var))
+        return SymbolicBool.from_z3(apply_smt(op, a.var, b.var))
 
     setup_binop(_, {ops.eq, ops.ne})
 
     def _(op: BinFn, a: SymbolicBool, b: bool):
-        return SymbolicInt(apply_smt(op, a.var, z3.BoolVal(b)))
+        return SymbolicInt.from_z3(apply_smt(op, a.var, z3.BoolVal(b)))
 
     setup_binop(_, _ARITHMETIC_OPS)
 
     def _(op: BinFn, a: bool, b: SymbolicBool):
-        return SymbolicInt(apply_smt(op, z3.BoolVal(a), b.var))
+        return SymbolicInt.from_z3(apply_smt(op, z3.BoolVal(a), b.var))
 
     setup_binop(_, _ARITHMETIC_OPS)
 
--- a/crosshair/libimpl/builtinslib.py
+++ b/crosshair/libimpl/builtinslib.py
@@ -167,6 +167,11 @@ class SymbolicValue(CrossHairValue):
     def from_name(cls, varname: str, typ: Type):
         raise CrosshairInternal(f"from_name not implemented in {cls}")
 
+    def __copy__(self):
+        if inside_realization():
+            return self.__ch_realize__()
+        return self.from_z3(self.var, self.python_type)
+
     def __deepcopy__(self, memo):
         if inside_realization():
             return self.__ch_realize__()
@@ -2182,7 +2187,7 @@ def make_optional_smt(smt_type):
     def make(creator, *type_args):
         space = context_statespace()
         varname, pytype = creator.varname, creator.pytype
-        ret = smt_type(creator.varname, pytype)
+        ret = smt_type.from_name(creator.varname, pytype)
 
         premature_stats, symbolic_stats = space.stats_lookahead()
         bad_iters = (
--- a/crosshair/libimpl/builtinslib.py
+++ b/crosshair/libimpl/builtinslib.py
@@ -167,7 +167,10 @@ class SymbolicValue(CrossHairValue):
         # override __new__, which means we can't just call
         # cls(var, typ).
         obj = object.__new__(cls)
-        obj.__init__(var, typ)
+        if typ is None:
+            obj.__init__(var)
+        else:
+            obj.__init__(var, typ)
         return obj
 
     @classmethod
@@ -1843,16 +1846,6 @@ class SymbolicCallable(SymbolicValue):
     def __init__(self, smtvar: Union[str, z3.ExprRef], typ: Type):
         SymbolicValue.__init__(self, smtvar, typ)
 
-    def __bool__(self):
-        return True
-
-    def __eq__(self, other):
-        return (self.var is other.var) if isinstance(other, SymbolicCallable) else False
-
-    def __hash__(self):
-        return id(self.var)
-
-    def __init_var__(self, typ: type, varname):
         type_args: Tuple[Any, ...] = type_args_of(self.python_type)
         if not type_args:
             type_args = (..., Any)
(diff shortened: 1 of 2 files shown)


Rik-de-Kort:
As noted in the issue comments, the pattern I'm trying out is using `__new__` to provide the required `type(list)([1, 2, 3])`-like functionality. The Symbolic itself, of course, is still a wrapper around a z3 value. We could consider removing __init__ entirely and just setting all the stuff on the from_z3 bit, or having that as __init__, now that I think about it.

I've only adjusted code in `builtinslib.py` so far. Currently everything is broken but I'm going to go through and see what I can fix next. That way I'll have a bit of intuition as to what to grep for down the line.

Here's a loose list of notes:
* `_ch_smt_sort` is provided as interface on `AtomicSymbolicValue`, override as appropriate in subclasses so the relevant `from_name` classmethod can be used.
* `__init_var__` was previously defined on a class instance, I'm trying to move it into the `from_name` classmethod

pschanely:
Pretty exciting. The explicitness of these factory functions will be nice to have independently.

You've probably noticed the python type argument and wondered why it's there. For atomic types, it is pointless, but it's useful to save for parameterized types like `List[int]`. All symbolics take it, just to have uniformity in the constructor.

You've may have also noticed that the `.var` member isn't always a z3 expression - sometimes it's a tuple of z3 expressions. (the typing annotations should really be changed to say this) This is common for symbolic containers, where it's useful to be able to independently reason about the size of the container.

Rik-de-Kort:
Thanks for the comment! I hadn't noticed that (so far I've only worked on the numerics).

Regarding the python types, figured that would be the case. Slightly related thing: the signature of `from_z3` requires it to be reimplemented everywhere, but I don't like this since it requires some hoop-jumping to in the end pass everything on to the initializer. I'm thinking of just having `None` as default argument and only passing it to the initializer if it's provided.

One snag I hit: in line 1253 of builtinslib.py, there's a comment "Copy on custom objects may require patched builtins." I'm not sure if that's implying we should patch copy.deepcopy, but [there appears to be some support](https://docs.python.org/3/library/copyreg.html#module-copyreg) for adjusting the methods by which something is copied. I think that would be a better option.

pschanely:
> Regarding the python types, figured that would be the case. Slightly related thing: the signature of `from_z3` requires it to be reimplemented everywhere, but I don't like this since it requires some hoop-jumping to in the end pass everything on to the initializer. I'm thinking of just having `None` as default argument and only passing it to the initializer if it's provided.

That sounds reasonable to me!

> One snag I hit: in line 1253 of builtinslib.py, there's a comment "Copy on custom objects may require patched builtins." I'm not sure if that's implying we should patch copy.deepcopy, but [there appears to be some support](https://docs.python.org/3/library/copyreg.html#module-copyreg) for adjusting the methods by which something is copied. I think that would be a better option.

Ah. You mean [core.py, here](https://github.com/pschanely/CrossHair/blob/c7b658cea619ed0cbbef1f65f53402d60d2d1917/crosshair/core.py#L1252). This point is a little more subtle - and it's referring to the idea that we should probably try to remove the surrounding NoTracing() block. These arguments might be symbolics, but might be concrete objects with symbolic members - this happens with arguments of user-defined classes, for example. When that user-defined class gets copied, it might customize its copy behavior (via `__copy__` or similar) and run arbitrary python. When that happens, we probably want to have tracing back on, so that if/when that code starts to do things with those symbolic members, all the appropriate patches are in place. 

(maybe I mentioned this before but system patches are registered with `register_patch` and can be enabled/disabled with `NoTracing()` and `ResumedTracing()`)

Anyway, I'm guessing you're having trouble with this copy when you add `__new__`? (perhaps something like [this SO question](https://stackoverflow.com/questions/59762417/unable-to-deepcopy-a-class-with-both-init-and-new-defined)) I am not super well versed in this space, but something like that copyreg module you mentioned might do the trick.

Rik-de-Kort:
As always in Python, it turns out there's a dunder method that does it. SymbolicValue had `__deepcopy__` already defined, but it was calling `copy` so I just defined `__copy__` to work with the new constructor methods. I  basically just copied `__deepcopy__`, which seems to make an awfully shallow copy?

Rik-de-Kort:
Quick unrelated question: in builtinslib.py on line 2161, why use `dict((v, k) for (k, vs) in thing for v in vs)` rather than `{v: k for (k, vs) in thing for v in vs}`?

pschanely:
> Quick unrelated question: in builtinslib.py on line 2161, why use `dict((v, k) for (k, vs) in thing for v in vs)` rather than `{v: k for (k, vs) in thing for v in vs}`?

"Old habits die hard." :) update at will!

Rik-de-Kort:
Can I get your feedback on the implementation of `from_name` for `SymbolicDict` and `SymbolicSet` ? I'm not sure about factoring out `_ch_key_types` and `_ch_val_types`, and calling `context_statespace` in the classmethod as it will get called again in the initializer. Maybe it's a better idea to refactor `__init__` to only set the variables to the right form, and then do the work for finding those variables in the constructors?

pschanely:
> Can I get your feedback on the implementation of `from_name` for `SymbolicDict` and `SymbolicSet` ? I'm not sure about factoring out `_ch_key_types` and `_ch_val_types`, and calling `context_statespace` in the classmethod as it will get called again in the initializer. Maybe it's a better idea to refactor `__init__` to only set the variables to the right form, and then do the work for finding those variables in the constructors?

I don't think the logic in `_ch_key_types` etc is expensive; seems reasonable to just call it twice. (though I probably also won't argue with you if you think it's cleaner to pass more arguments into `__init__`)

Having the StateSpace as a member on the Sybolic* classes is a leftover from an older time (hopefully, it's always the same as `context_statespace()`). Ultimately, I think it should be removed (even though accessing the thread local presumably has a slightly higher cost). At any rate, I think that means you shouldn't worry about adding calls for `context_statespace()` - eventually, we'll just always be doing that.

Rik-de-Kort:
Little todo: unify the `_ch_pytypes`-family of functions. Currently getting my head around the internals of symbolic collections.

Libimpl in __init__ of SymbolicObject, is that not meant to be `SymbolicType(smtvar, typ)` instead of `SymbolicType(smtvar, type)`?

pschanely:
A broad comment: this is an ambitious endeavor. Feel free to get part of the way and have me wrap it up.

> Little todo: unify the `_ch_pytypes`-family of functions. Currently getting my head around the internals of symbolic collections.

There is quite a bit of complexity in there so that both reasoning about sizes/lengths and iteration both work well.

> Another one: should probably add custom from_z3 to SymbolicList since it has that ShellMutableSequence wrapper.

+1

> in builtinslib.py in `__init__` of SymbolicObject (line 1744 on current master), is that not meant to be `SymbolicType(smtvar, typ)` instead of `SymbolicType(smtvar, type)`?

This is a little esoteric. `SymbolicType` represents an unknown/symbolic type. `SymbolicObject` represents an unknown/symbolic instance.

`typ` arguments to `__init__` should correspond to python type for the symbolic; therefore, we expect to see `type` for `SymbolicType` and `object` for `SymbolicObject`, respectively.

Instead of `SymbolicType(smtvar, type)`, we might have considered `SymbolicType(smtvar, Type[typ])`. But I think, in practice, `SymbolicObject.__init__` always gets called with `object` as `typ`, which would yield `Type[object]`, i.e. just `type`. If we were creating an object with a known type, we would just be calling the function for that more specific type, rather than `SymbolicObject`.

> Also am going to have to look into how CrossHair works on classes, looks like. Any pointers?

I wish I had real docs here. :)  We have two tactics for classes; see `proxy_class_as_concrete` and `proxy_class_as_masquerade`.

`proxy_class_as_concrete` tries to create a real, regular instance, passing symbolics symbolics to the constructor and overwriting typed members with symbolics.

`proxy_class_as_masquerade` creates a proxy class that extends the target type, hoping to avoid constructor logic, and assigns typed members with symbolics.

Neither of these approaches create new symbolic values, so you won't need to do anything special for the task at hand ... I think. :)

Rik-de-Kort:
Not dead yet! Just a bit short on time. :) Need to get SymbolicCallable properly factored.

Rik-de-Kort:
We now have some passing tests!!

Some comments:
- The first test in bugs_detected was giving the list `[object(), 0, 0]`, rather than `[0, 0]`, which mysteriously disappeared after I fixed up `SymbolicCallable`. I was expecting it would have something to do with `SymbolicList` instead. 
- Currently there's some duplicated logic in `__init__` and `from_name` in `SymbolicCallable`. Needs a bit of cleanup.
- I need to consider somewhat carefully the impact of typ=None as default argument. While I *think* it usually works, I have no guarantee and Mypy is rightfully not happy about it.

pschanely:
> We now have some passing tests!! Not the whole test suite but I think this is something like "initial functionality".

Nice!

> * The first test in bugs_detected was giving the list `[object(), 0, 0]`, rather than `[0, 0]`, which mysteriously disappeared after I fixed up `SymbolicCallable`. I was expecting it would have something to do with `SymbolicList` instead.

This might be another gotcha - the counterexamples [aren't stable](https://github.com/pschanely/CrossHair/blob/9a7185ef2365a4319bfb046a5450f7c4761751c8/crosshair/check_examples_test.py#L209), which leads to some confusing behavior in the bugs_detected tests, where any one failure causes many things to look like they are failing too.


> * Currently there's some duplicated logic in `__init__` and `from_name` in `SymbolicCallable`. Needs a bit of cleanup.
> * I need to consider somewhat carefully the impact of typ=None as default argument. While I _think_ it usually works, I have no guarantee and Mypy is rightfully not happy about it.
> * I need to implement `__new__` for collection types. This is probably what is currently causing `check_iter` to fail.

I am underwater with a lot of my "day job" this week, but I hope to have some time this weekend - I'll pull this work locally and see whether I have any useful insights. More soon!

Rik-de-Kort:
I don't think we're in a rush here, so take it easy if you need to. Though I think some review might be very helpful at this point! 

Taking stock, I currently have the following outstanding points:
- The presence of the `None` default argument on `from_z3`. All the initializers take care of it as well, though if I take out the `None` handling in `from_z3` on `SymbolicValue`, the csv example fails. Responsibility for handling this `None` should be in one place, which might have to be the initializer.
- There are methods for figuring out the crosshair and python types from the passed in `typ` argument for the collection types and `SymbolicCallable`. These get called multiple times on each value construction, which doesn't look great.
- I need to implement `__new__` for the collection types. I *could* implement it for callables as well, but that seems farfetched and I would rather write an error that says "pls no".

Below this is my stream-of-consciousness writing to come to the conclusions above. Feel free to skip.

There are three ways to construct a symbolic value: `from_name`, `from_z3`, and `__new__`. `from_name` has signature `from_name(cls, varname: str, typ: Type=None)`, `from_z3` has signature `from_z3(cls, var, typ: Type=None)`. `__new__` has signature equivalent to whatever is used in Python itself.

At the bottom of the hierarchy is `SymbolicValue`. It inherits from `CrossHairValue`, which only provides an override for deepcopy. It implements `from_z3` and `from_name` to provide "sensible defaults":
- `from_name` implements a `z3.Const` pattern, relying on `_ch_smt_sort()` to provide the type. This suffices to implement `from_name` for atomic types, i.e. bools, ints, floats, strings.
- `from_z3` takes a z3-value and simply passes that into the intializer, in a way that bypasses `__new__` in case we have overridden it.

`SymbolicBool` is the next one with its own constructor logic. Here we override `__new__` to provide constructor compatibility. `SymbolicBool` also implements `_ch_smt_sort` to provide the type used in `from_name`.  
`SymbolicInt`, `SymbolicFloat`, and `SymbolicString` are very similar.

One thing I did is to accept `None` as default argument for `typ` in the initializers. The reasoning behind this is that I can just call the initializer from `from_z3` where this variable might be `None`. This is not reflected in the implementation of `SymbolicValue`, though. Trying to take it out leads to the csv example failing -> will have to check why.

Next up are the collection types. This starts with `SymbolicDictOrSet`. Sets in our case are modelled as dicts with only a key type. To this end there is a method `_ch_key_types` that infers the type of the keys from the passed in Python type. `SymbolicDict` additionally has a method `_ch_val_types` that infers the type of the values from the passed in Python type. For `SymbolicArrayBasedUniformTuple` (which backs `SymbolicList`), we have `_ch_types`. These methods are called in the initializers.  
However, these methods are *also* called in the `from_name` constructors. This is necessary because we have to create a z3 object of the right type. I haven't implemented it yet, but they will also be called in `__new__`, and potentially in `from_z3`, which needs to be overridden as the `z3.Const` pattern is wrong for these types. If we're calling from all the constructors, it seems reasonable to just pass the result into the initializer rather than doing that work again. Though then the signature of `__init__` is not so great.  

Then we have `SymbolicType`, which I have done very little for. It's treated basically like `SymbolicBool`.

Finally, `SymbolicCallable` implements a similar pattern as the collections with `_ch_func_types`.

crosshair/libimpl/builtinslib.py
```diff
@@ -811,9 +828,13 @@ def __round__(self, ndigits=None):
...
 class SymbolicInt(AtomicSymbolicValue, SymbolicIntable):
-    def __init__(self, smtvar: Union[str, z3.ExprRef], typ: Type = int):
+    def __init__(self, var: z3.ExprRef, typ: Type = None):
+        typ = int if typ is None else typ
         assert typ == int
-        SymbolicIntable.__init__(self, smtvar, typ)
+        SymbolicIntable.__init__(self, var, typ)
+
+    def __new__(cls, value: int):
+        return cls.from_z3(z3.IntVal(value))
```
pschanely, line 837: I'll note that the native python constructors we're emulating can be pretty fancy, e.g. `int("8a1f", 16)`.
What happens if we started with this? (on AtomicSymbolicValue maybe):
```
def __new__(cls, *a):
  return cls._pytype()(*a)
```
This is worse than your solution because it will realize all symbolic state in the arguments. (or at least I think it will; if not, we will need to call `realize()` on the arguments explicitly) But perhaps it's easier to just try to get the behavior right first. We can optimize for preserving symbolics later, where we think we really want it.

crosshair/libimpl/builtinslib.py
```diff
@@ -152,18 +152,39 @@ def smt_coerce(val: Any) -> z3.ExprRef:
...
+        # support constructor syntax. To do this we
+        # override __new__, which means we can't just call
+        # cls(var, typ).
+        obj = object.__new__(cls)
+        if typ is None:
+            obj.__init__(var)
         else:
-            self.var = smtvar
-            # TODO test that smtvar's sort matches expected?
+            obj.__init__(var, typ)
```
pschanely, line 173: This is the None handling that you were referring to, breaking the csv example, yes?
This `if` statement looks reasonable and necessary to me. You might be thinking that passing `None` into an parameter with a default value causes Python to start using the default value: nope, it will still just receive `None`.

pschanely:
I left a comment above with a big, wacky idea for our implementations of `__new__`: we could just call the real type's constructor and return that. (I believe `__new__` isn't required to return an object of the same type that it was called on!) I'm not clear on whether that approach would invalidate or change some of the other discussion here.

> * The presence of the `None` default argument on `from_z3`. All the initializers take care of it as well, though if I take out the `None` handling in `from_z3` on `SymbolicValue`, the csv example fails. Responsibility for handling this `None` should be in one place, which might have to be the initializer.

See my in-code comment in SymbolicValue.from_z3 (I'm not sure whether I am interpreting you correctly here though!).

> * There are methods for figuring out the crosshair and python types from the passed in `typ` argument for the collection types and `SymbolicCallable`. These get called multiple times on each value construction, which doesn't look great.

FWIW, I am not overly concerned with having to call these multiple times.

> * I need to implement `__new__` for the collection types. I _could_ implement it for callables as well, but that seems farfetched and I would rather write an error that says "pls no".

It seems perfectly reasonable to `raise CrossharUnsupported` here, since it doesn't work right now either.

> * I need to implement `from_z3` for the collection types. The implementation in `SymbolicValue` is wrong for them.

+1

Rik-de-Kort:
Thanks for the review! Will dig into it when I have the time. :)

crosshair/libimpl/builtinslib.py
Rik-de-Kort: The trouble is I'm doing None-handling in both `from_z3` and `__init__`. Ideally I'd keep it in one spot so I don't forget to implement it in one and then not the other.

Rik-de-Kort:
@pschanely you mentioned cleanup in this PR that might still be nice. Do you mean the refactoring into `from_z3` and `from_name` classmethods, or the splitting up of `_ch_val_types` and friends? Or both?

pschanely:
Potentially both! WDYT?

* I still like the explicitness of the `from_z3` and `from_name` classmethods.
* And IIRC the work involving `_ch_val_types` let us get rid of the `__init_var__` methods. I was never really happy with those.

That said, I think the codebase has moved some since you were last looking at this, so perhaps the cost/benefit tradeoff is less attractive now - unsure!

Rik-de-Kort:
Both is fine though that might mean having to rethink the exact functionality of the init method. What I'll do is I'll take this work as inspiration and port it over to a new branch, see where we end up.

PR #122, opened 2021-10-12
Add pre-commit

nicpayne713:
For review... this is an initial stab at migrating your `precommit.py` file to the `pre-commit` git hook. It can be run by first installing `pre-commit` with `pip install pre-commit` then running `pre-commit install` and then you can run it with `pre-commit run` and since it's a git hook it will run on commit - which can be bypassed with `--no-verify`.

Pre-commit docs are pretty solid, and I just thought, in the spirit of hacktoberfest to see if you might like to migrate.

Note: No doctest is here, or pytest necessarily but I think we could work them in if desired.

Reasons: the config is much more extensible and more easily worked with than a custom .py file and there's wide support for numerous hooks.

Let me know what you think!

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,6 @@
+# flake8 config
+
+[flake8]
+max-line-length = 79
+max-complexity = 18
+select= E9,F63,F7,F82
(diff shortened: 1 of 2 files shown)


pschanely:
At a high level, this all sounds good - thanks for investigating!
I'm happy to provide feedback along the way too.

pschanely:
I'm going to close this due to inactivity. I'm still open to the idea, though, so if anyone wants to continue this work, re-open at will!

Issue #119, opened 2021-10-08
Use more flake8 


pschanely:
Right now (in [precommit.py](https://github.com/pschanely/CrossHair/blob/98bc53798467f03c71ad63f041c7177894c86aa8/precommit.py#L95)) we only enable a few checks (E9, F63, F7, F82).

Looks like there are a lot of other useful ones in there; investigate moving to a set of blocklisted checks.

nicpayne713:
Hello, I briefly looked over your repo and I'm unfamiliar with using a .py file for precommit... is there any reason you don't configure pre-commit with a `.pre-commit-config.yaml` and then run `pre-commit run all` or have a hook that runs pre-commit for you? A custom .py file feels like a lot of unnecessary customization at first glance... I'm not an expert here by the way, so if there's good reason I'm more than happy to be wrong. If though you'd like to transition to a pre-commit config I'd be more than happy to help close this issue with one. 
Cheers.

nicpayne713:
Hey @pschanely as we are discussing pre-commit elsewhere, what did you have in mind for extending your current .py file? I'd be happy to help move that along and make it more usable regardless of your decision to migrate to pre-commit or not

nicpayne713:
Cool I'll work on some small edits on a branch for this issue - will include `isort` by the way

PR #130, opened 2021-10-18
Remove unused imports and variables

orsinium:
1. Remove unused imports using [autoflake](https://github.com/myint/autoflake) and [unimport](https://github.com/hakancelik96/unimport).
2. Remove unused variables using autoflake
2. Expand starred imports using autoflake
2. Add `__all__` into facades.

The first step to #119

### DIFF
--- a/crosshair/fnutil.py
+++ b/crosshair/fnutil.py
@@ -21,7 +21,6 @@ from typing import (
     Dict,
     Iterable,
     Optional,
-    Protocol,
     Tuple,
     Type,
     Union,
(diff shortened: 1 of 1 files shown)


pschanely:
> Another note: I run tests in parallel using pytest-xdist, works like a charm. Probably, it makes sense to do the same on CI, depends on how many cores you have there.

Realized I never responded to this! Indeed, I sometimes also use `pytest-xdist` when running locally. (as well as `pytest-random-order`, sadly the two plugins don't work together)

For CI, I am uncertain whether it's better to keep things maximally deterministic. Perhaps at a minimum, I'll make xdist part of the dev dependencies, and have precommit use it by default.

orsinium:
Try [pytest-randomly](https://github.com/pytest-dev/pytest-randomly), it also randomizes the order of the tests but also does support pytest-xdist.

PR #150, opened 2021-12-17
WIP: Add coverage to `check` operation

samuelchassot:
This PR is for discussion purposes for now :)
@gondolav

### DIFF
--- a/crosshair/core.py
+++ b/crosshair/core.py
@@ -47,6 +47,8 @@ from typing import (
 )
 import typing
 
+from crosshair.util import CoverageResult, measure_fn_coverage
+
 import typing_inspect  # type: ignore
 import z3  # type: ignore
 
@@ -637,7 +639,11 @@ class ConditionCheckable(Checkable):
                 )
             ]
         elif analysis.verification_status is VerificationStatus.CONFIRMED:
-            message = "Confirmed over all paths."
+            if analysis.coverage_result:
+                message = f"Confirmed over all paths with coverage = {analysis.coverage_result.opcode_coverage}."
+            else:
+                message = "Confirmed over all paths."
+
             analysis.messages = [
                 AnalysisMessage(
                     MessageType.CONFIRMED,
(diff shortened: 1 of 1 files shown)


samuelchassot:
We however found something interesting (that you might find interesting too :) ):
Python compiler adds two opcodes (`load_const None` and `return_value`) at the end of functions (among other structures).
For example, this function:
```
def f(a: int) -> int:
    """
    pre: a < 0
    post: __return__ < 0
    """
    b = a * 3
    if b == -3:
        return -1
    else:
        return b
```
produces this bytecode:

```
  6           0 LOAD_FAST                0 (a)
              2 LOAD_CONST               1 (3)
              4 BINARY_MULTIPLY
              6 STORE_FAST               1 (b)

  7           8 LOAD_FAST                1 (b)
             10 LOAD_CONST               2 (-3)
             12 COMPARE_OP               2 (==)
             14 POP_JUMP_IF_FALSE       20

  8          16 LOAD_CONST               3 (-1)
             18 RETURN_VALUE

 10     >>   20 LOAD_FAST                1 (b)
             22 RETURN_VALUE
             24 LOAD_CONST               4 (None)
             26 RETURN_VALUE
```

So, the 2 last opcodes are unreachable, which means that the coverage output by Crosshair is < 1.

Moreover, it appears that sometimes (we couldn't figure out when exactly for now), these two opcodes are not there (probably some optimisations going on).

pschanely:
Also beginning to wonder whether I've led you down a fruitless path. Even accurate bytecode-level coverage may not be sufficient to detect nondeterministic behaviors, e.g. it's easy to get full coverage on this variant of the `time()` example:

```
import time
def main():
    '''post: __return__ != 0'''
    x = int(time.time())
    return x % 235432
```

pschanely:
I'm going to close this due to inactivity. Bytecode-level coverage seems challenging, but the general concept of having CrossHair report coverage seems valuable nevertheless. Re-open at will!

PR #160, opened 2022-03-28
External contract registration

pschanely:


crosshair/main.py
```diff
@@ -94,6 +95,11 @@ def command_line_parser() -> argparse.ArgumentParser:
         metavar="FLOAT",
         help="Maximum seconds to spend checking execution paths for one condition",
     )
+    common.add_argument(
+        "--contract_file",
```
pschanely, line 99: One might want multiple files too?

It's interesting that a `contract_file` is very similar to a plugin - they are both just python files that we run at startup. What do you think about naming this `--extra_plugins`? In theory someone might want to conditionally apply `register_class()` via the command line, and conversely someone might use `register_contract()` via a plugin.

crosshair/register_contract.py
```diff
@@ -0,0 +1,104 @@
...
+from crosshair.stubs_parser import signature_from_stubs
+from crosshair.util import debug
+
+
+class ContractRegistrationError(Exception):
+    pass
+
+
+@dataclass
+class Contract:
```
pschanely, line 15: Can we name this something like `ContractOverride`? Just to be extra clear that this isn't the class CrossHair directly uses to check contracts.

pschanely (commented):
I created a draft pull request just to make it easier to give you some early feedback. I come from a pretty intensive code review culture - apologies in advance if it seems picky. A lot of it is just naming suggestions etc, but that kind of stuff is always better to get out of the way early. Finally, don't hesitate to push back on any suggestion of mine.

pschanely:
> I created a draft pull request just to make it easier to give you some early feedback. I come from a pretty intensive code review culture - apologies in advance if it seems picky. A lot of it is just naming suggestions etc, but that kind of stuff is always better to get out of the way early. Finally, don't hesitate to push back on any suggestion of mine.

Oh my goodness. Github just ate ~5 of my comments so there's much less here than there should be. Will have to try a remember/rewrite them tomorrow!😭

crosshair/register_contract.py
lmontand: I totally agree with that, I will rename it soon!

lmontand:
> I created a draft pull request just to make it easier to give you some early feedback. I come from a pretty intensive code review culture - apologies in advance if it seems picky. A lot of it is just naming suggestions etc, but that kind of stuff is always better to get out of the way early. Finally, don't hesitate to push back on any suggestion of mine.

Perfect, I agree that early code review is a good idea and making a PR is the best way if discussing it! Being rigorous is a good thing!
I have indeed been hesitating for many of the variable names, so your feedback on those is welcomed!

crosshair/main.py
lmontand: Allowing multiple files is a good idea!

I have indeed been hesitating a lot on this name. I like the name `--extra_plugin`. The drawback is that it does not specify we are talking about contract registration. But at the same time, it's true that it might not only concern registering contracts; it could be used as well for the existing [plugins](https://crosshair.readthedocs.io/en/latest/plugins.html).

pschanely (commented):
Ok! I re-typed my lost comments from yesterday! Overall, this is looking really great.

crosshair/main.py
pschanely: > I have indeed been hesitating a lot on this name. I like the name `--extra_plugin`. The drawback is that it does not specify we are talking about contract registration. But at the same time, it's true that it might not only concern registering contracts; it could be used as well for the existing [plugins](https://crosshair.readthedocs.io/en/latest/plugins.html).

Yup! I believe it'll still make sense to have "how-to" documentation specific to contract registration, but at the level of the command line interface, I'm inclined to think of all these things as plugins.

crosshair/stubs_parser.py
```diff
@@ -0,0 +1,289 @@
...
+    stub_file = get_stub_file(fn.__module__, search_context=search_context)
+    module = get_stub_ast(fn.__module__, search_context=search_context)
+    if not stub_file or not module or not isinstance(module, ast.Module):
+        debug("No stub found for module", fn.__module__)
+        return None
+    glo = globals().copy()
+    return _sig_from_ast(module.body, path_in_module, stub_file.read_text(), glo)
+
+
+def _sig_from_ast(
```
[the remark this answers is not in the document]
lmontand: There are similarities: For example `typeshed_client.get_stub_names("random")` returns the AST of all objects in the `random` module. This way one can easily find the `Random` object. You can even go one step further using `resolver = typeshed_client.Resolver()` and `resolver.get_fully_qualified_name("random.Random")`.
I first tried it this way, but the problem is that some parameter annotations might not be defined in the current namespace. So evaluating the function type annotations might fail. This is the reason why I go through the entire AST of the module to find the function and I execute the import statements and `TypeVar` definitions found on the way to have a namespace as close as possible as what is required.
I'm not fully satisfied of doing it this way, but I did not find a better way. I think that since stub files are not supposed to be used at runtime, there is no very clean way of getting the signature of a function.

crosshair/stubs_parser.py
```diff
@@ -0,0 +1,289 @@
...
+        if sys.version_info < (3, 9):
+            annotation = _rewrite_with_typing_types(annotation, glo)
+        try:
+            return eval(annotation, glo)
+        except Exception as e:
+            debug("Not able to parse annotation:", annotation, "Error:", e)
+    return Parameter.empty
+
+
+def _rewrite_with_union(s: str) -> str:
```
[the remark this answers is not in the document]
lmontand: Thank you for the hint, I have never touched to any AST before, so I didn't know this. Good to know for next time!

setup.py
```diff
@@ -40,6 +40,7 @@
         "typing_extensions>=3.10.0",
         "z3-solver==4.8.9.0",
         "importlib_metadata>=4.0.0",
+        "typeshed-client==2.0.4",
```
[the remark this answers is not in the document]
lmontand: I think we can. The main risk of doing this is that the stubs syntax might change once more with Python 3.11 and not be backward compatible, but I don't think that the changes will be that big. As far as I know, the only new change planned so far is [PEP 673](https://peps.python.org/pep-0673/).
And in the worst case, annotations the parser is not able to parse are replaced by `Parameter.empty`.
The best thing would be to follow updates of `typeshed` to see if there are problematic changes which would imply adapting the parser.
I will change this to `>=2.0.4` in the following days (unless this made you change your mind ;-) )

crosshair/condition_parser.py
```diff
@@ -1161,12 +1168,106 @@ def get_class_invariants(self, cls: type) -> List[ConditionExpr]:
...
+                    f"manually registered precondition for {fn.__name__}",
+                )
+            )
+        if contract.post:
+            post_cond = contract.post
+
+            def post_eval(orig_kwargs: Mapping) -> bool:
+                kwargs = dict(orig_kwargs)
+                kwargs["result"] = kwargs["__return__"]
+                kwargs["OLD"] = kwargs["__old__"]
```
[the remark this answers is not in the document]
lmontand: I had the same reasoning last week. I think I will roll back to using `__return__` and `__old__`. There is no real reason of using `iContract`'s syntax more than PEP 316. Do you think we should also allow `_` in addition to `__return__` for the return value?

crosshair/stubs_parser.py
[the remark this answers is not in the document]
pschanely: Ah, so you already tried it; that's good enough for me!

crosshair/condition_parser.py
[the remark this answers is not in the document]
pschanely: > Do you think we should also allow `_` in addition to `__return__` for the return value?

I feel less strongly about this, either way. Perhaps just `__return__`. Programmatic interfaces don't need the same niceties, because people can always build a layer on top of it.

setup.py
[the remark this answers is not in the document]
pschanely: Heh, your thoughts have given me some pause.
But sadly, I think it's unlikely that I'll even remember to increment the version if we pin it.

I think GitHub CI will run with the current versions of all dependencies, so perhaps a few good tests could catch some kinds of regressions.

crosshair/register_contract_test.py
```diff
@@ -22,12 +26,110 @@ def test_register_malformed_contract():
...
+        """
+        pre: x <= 10
+        post: _ >= x
+        """
+        return randint(x, 10)
+
+    register_contract(
+        Random.randint,
+        pre=lambda a, b: a <= b,
+        post=lambda __return__, a, b: a <= __return__ and __return__ <= b,
```
pschanely, line 52: BTW, Python will let you chain comparison operators, so `a <= __return__ <= b` should be equivalent.

crosshair/stubs_parser.py
```diff
@@ -104,28 +126,36 @@ def _sig_from_ast(
...
         if n.name == "_typeshed":
             return
+    # Replace imports from `_typeshed` by their equivalent
     module = getattr(imp, "module", None)
     if module == "_typeshed":
+        for n in imp.names:
+            name = n.name
+            if name in _REPLACE_TYPESHED:
+                new_module, replace = _REPLACE_TYPESHED[name]
+                exec("from " + new_module + " import " + replace + " as " + name, glo)
```
pschanely, line 155: Consider [importlib.import_module](https://docs.python.org/3/library/importlib.html#importlib.import_module)  (+getattr) for this.

pschanely (commented):
Still looking great. Just left a few more tiny notes.

crosshair/register_contract_test.py
lmontand: Great, thank you for letting me know! I'll update it.

crosshair/stubs_parser.py
lmontand: I agree this would be better this way. I'll update it.

pschanely (commented):
Looking great.
Just one thought above; otherwise, I'm good to merge if you are!

doc/source/plugins.rst
```diff
@@ -71,3 +78,118 @@ distribution with a key of ``crosshair.plugin``, like so:
...
+function to register **all** functions of a module. Simply pass the modules you want to
+register as an argument to the function. At runtime, CrossHair will register all
+functions encountered which belong to this module, assuming no pre- or postcondition.
+If you want to specify a pre- or postcondition, use the ``register_contract`` function
+seen above.
+
+.. note::
+    Functions ``__init__``, ``__init_subclass__`` and ``__new__`` are not concerned by
+    module registration. Also note that registering a whole module might be too much in
+    some cases and you might fallback to register individual functions instead.
```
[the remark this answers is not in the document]
lmontand: Good thought! Indeed, I don't have a lot of experience about this feature so far. I only tried it on `time`, `random` and `numpy` modules. It worked for the two firsts, but `numpy` seems to be too complex. For now, I don't know if we will use it for SVSHI, but I will be working on it next week (I will add contract registration to SVSHI where it has non-determinism or side-effects).
So, I will remove this from the doc and mention in the docstring that this is an experimental feature.

Also, did you think of better examples than `numpy.random.randint` and `random.randint` for the doc?

PR #181, opened 2022-10-20
Switch to use pre-commit config YAML file

### LINKED ISSUE #121: Consider other precommit approaches
Hello, I briefly looked over your repo and I'm unfamiliar with using a .py file for precommit... is there any reason you don't configure pre-commit with a `.pre-commit-config.yaml` and then run `pre-commit run all` or have a hook that runs pre-commit for you? A custom .py file feels like a lot of unnecessary customization at first glance... I'm not an expert here by the way, so if there's good reason I'm more than happy to be wrong. If though you'd like to transition to a pre-commit config I'd be more than happy to help close this issue with one. 
Cheers.

_Originally posted by @nicpayne713 in https://github.com/pschanely/CrossHair/issues/119#issuecomment-939202603_

tekktrik:
Resolves #121 by swapping `precommit.py` for the standard `.pre-commit-config.yaml` file.  I tried my best to replicate the existing actions performed by the old file, and also tried to fix any errors that popped out.  There are a few remaining from some of the hooks, so please feel free to push any fixes for them directly to this PR.

Setting this as a draft for now, if everything looks good I can undraft it.

Let me know if you have any questions, or if there's anything I can explain!


.pre-commit-config.yaml
```diff
@@ -0,0 +1,52 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.2.0
+    hooks:
+      - id: check-yaml
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+  - repo: https://github.com/python/black
+    rev: 22.3.0
```
pschanely, line 9: Does pre-commit auto-install hooks? (thanks in advance for dealing with my ignorance about pre-commit!) If not, let's pin the dev dependencies for each of these in setup.py. If it does auto-install dependencies, let's remove those dependencies from setup.py.

And I guess we should add pre-commit itself as a dev dependency in setup.py?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,52 @@
...
+      - id: __init__ + setup.py
+        name: Check __init__ matches setup.py
+        entry: python crosshair/tools/check_init_and_setup_coincide.py
+        language: system
+      - id: help in doc
+        name: Check help in doc
+        entry: python crosshair/tools/check_help_in_doc.py --overwrite
+        language: system
+        pass_filenames: false
+      - id: pytest
```
pschanely, line 48: I'm worried that the full test suite may be too much to run per-commit. Do you have a suggestion for what's appropriate here? (as a follow-up, I can work on some sort of "smoke test" phase)

Perhaps we put this in the "push" or "manual" stage?: https://pre-commit.com/#confining-hooks-to-run-at-certain-stages

We DO want all the tests to run in the github action, of course.

conftest.py
```diff
@@ -0,0 +1,9 @@
+"""
+Configure settings before running pytest
+
+In this case, we set PYTHONHASHSEED to 0 for uniformity across test runs.
+"""
+
+import os
+
+os.environ["PYTHONHASHSEED"] = "0"
```
pschanely, line 9: Sadly, I don't believe this changes the actual value used by the interpreter (the actual hash must remain constant for the entire process). So, although it satisfies the tests, it doesn't accomplish what I'm looking for.

Honestly, I'm not sure how to do this in pre-commit. It looks like the pre-commit folks [don't want to support env variables](https://github.com/pre-commit/pre-commit/issues/1157) in the hook configuration, and I don't want to introduce something that will break on Windows. (some tests don't pass on Windows right now, but I'm actively working to fix that)

Python has a command-line option to force a random seed, but no option to un-randomize it.

So ... maybe we need to resort to using a shell Python process to set the variable? 😭 
```
python -c 'import os, subprocess; subprocess.run("python -m pytest", shell=True, env={**os.environ, "PYTHONHASHSEED":"0"})'
```

pschanely (changes requested):
After seeing this for real, I'm even more excited about it. Thank you!

Maybe take a look at some of my suggestions here, pull it out of draft mode, and I'll approve CI runs, so we can see it in action!

conftest.py
pschanely: Also! Add yourself to the contributor list in `contributing.rst` and link to whatever you like. 😄

.pre-commit-config.yaml
tekktrik: Yup, it will install the specified hook at the given rev, so it will install when it first runs.  The great thing is that if you change the rev it will install the new one too.  Since it's all in a virtual environment pre-commit uses, you could remove them as dependencies in `setup.py` if you wanted, and pre-commit will manage it.

Note that this is only true for non-local hooks (local ones are at the end, and those run on the given system so things like `pytest` would still be needed).

.pre-commit-config.yaml
tekktrik: That's a fair concern. I didn't actually know about the `stages` property, but that's a good idea!  I can add the testing to only run on either pushes or manual, if you have a preference.  Let me know!

`pre-commit run --all-files` should still run all the hooks, if I understand correctly, so I think that would run even ones with stages.

conftest.py
tekktrik: So `conftest.py` is a special file for `pytest` that runs before any tests are run, and when I ran the `pytest` locally before and after adding this, I did see the pytest that checks the seed stop failing, for what it's worth.  It should also not be permanent, as adding things to `os.environ` only last for the duration of the Python interpreter process, if I understand correctly.  Let me know if I'm misunderstanding the issue though.

tekktrik: And I'll go ahead and add myself, and the other changes when those are decided!

tekktrik: I also meant to mention that this pytest hook runs locally, so using `conftest.py` to set the environment should definitely work!

pschanely: Yeah, so changing `os.environ` will trick my silly test into passing, but unfortunately it doesn't change the underlying seed used by python. Check this out:
```
$ PYTHONHASHSEED=0 python -c 'print(hash("foo"))'
-2457967226571033580
$ PYTHONHASHSEED=0 python -c 'print(hash("foo"))'
-2457967226571033580
$ python -c 'import os; os.environ["PYTHONHASHSEED"] = "0";  print(hash("foo"))'
3608991658523737785
$ python -c 'import os; os.environ["PYTHONHASHSEED"] = "0";  print(hash("foo"))'
-5911599181240111572
```
When we set the environment variable before python starts, the behavior of hash is consistent across runs. If we set it in `os.environ`, it changes across runs. (b/c by then the Python interpreter doesn't care - it's already picked its seed)

.pre-commit-config.yaml
pschanely: Great. Let's do "manual" for now; we can change if needed later.

.pre-commit-config.yaml
pschanely: > Since it's all in a virtual environment pre-commit uses, you could remove them as dependencies in `setup.py` if you wanted, and pre-commit will manage it.

Fantastic; let's do that.

conftest.py
tekktrik: So I think we can get around it be using this entry:

```
entry: env PYTHONHASHSEED=0 python -m pytest
```

Looks like that's the reported fix, and sets it right before calling Python and only for that specific spin of the interpreter based on my local testing.

pschanely: > entry: env PYTHONHASHSEED=0 python -m pytest

My only reservation for this solution is that it feels like it won't work on windows (unless it's cygwin python perhaps?). I could test this tomorrow. I've got other problems with windows, so it's not the biggest deal, but it's one more thing that we'll have to unwind eventually. Anyway, that's why I was proposing that horrendous `python -c` construction earlier. What do you think?

tekktrik: Tested on my Windows machine just now, and getting the same result using `cmd`:

```shell
C:\Users\adelaney>python
Python 3.9.7 (tags/v3.9.7:1016ef3, Aug 30 2021, 20:19:38) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.environ["PYTHONHASHSEED"]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\adelaney\AppData\Local\Programs\Python\Python39\lib\os.py", line 679, in __getitem__
    raise KeyError(key) from None
KeyError: 'PYTHONHASHSEED'
>>> quit()

C:\Users\adelaney>env PYTHONHASHSEED=0 python
Python 3.9.7 (tags/v3.9.7:1016ef3, Aug 30 2021, 20:19:38) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.environ["PYTHONHASHSEED"]
'0'
```

I'm open to either solution though, but I agree that avoiding the awful `python -c` construction would be great.

pschanely: Oh wow. I stand corrected! Great, let's go with the `env` approach then!

tekktrik: Sounds like a plan!

tekktrik:
Whoops, I forgot to add `pre-commit` as a dependency!

tekktrik:
I pinned pre-commit with compatibility release notation to 2.20 (latest), but it could almost certainly be relaxed if there are any use cases requiring it.

tekktrik:
It looks like even `pre-commit run --all-files` alone wasn't running the `pytest` hook, so I patched that by adding `--hook-stage manual` to trigger it as well.

tekktrik:
Here are the failures that I see so far:

* There are two errors from `pydocstyle` for missing argument descriptions.  If you want to add those directly to this PR, it's open for changes to maintainers.
* It also looks like there's a `doctest` error but I'm having a harder time understanding *where* the error is there.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,53 @@
...
+    rev: v0.982
+    hooks:
+      - id: mypy
+  - repo: local
+    hooks:
+      - id: doctest
+        name: Run doctest
+        entry: python -m doctest
+        language: system
+        files: ".+.rst"
```
tekktrik, line 38: I also realize I'm not testing `README.md` like before, I'll add that back in.

tekktrik:
I fixed the `doctest` issue with that fix as well, but I guess one of tests in the `pytest` suite requires `mypy` and so that one failed.  Not sure if that means you want to keep `mypy` as a dev dependency, or whether the test should be changed.  Let me know!

pschanely:
> * There are two errors from `pydocstyle` for missing argument descriptions.  If you want to add those directly to this PR, it's open for changes to maintainers.

It's a little odd that [this exclusion](https://github.com/pschanely/CrossHair/blob/afedcd791799806bfd618bbc50fb9b602bd76128/setup.cfg#L30) in setup.cfg doesn't skip it; I think it has to do with pre-commit passing filenames directly. Anyway! Yup, I'll just fix up the two errors instead of thinking about it.

tekktrik:
I should be able to mimic that exclusion in the hook, I think, if you want to go that route as well.

pschanely (approved):

pschanely:
> Yup, tested locally, I can exlude it by passing `--match='(?!datetimelib).*(?<!_test).py$'` as an arg to the hook.

Thanks for investigating! I just pushed something to resolve the two errors in datetimelib.py - a reasonable thing to do anyway.

pschanely:
Checks look good, so I've merged. Thank you -- I'm pretty excited about the other hooks we can trivially add now!

I don't think I should have to, but adding the "hacktoberfest-accepted" label to make extra sure it counts :smile:

tekktrik:
Thanks so much!  I always love working on repo infrastructure.  Feel free to reach out if anything like that comes up!

PR #185, opened 2022-11-13
Save Hypothesis buffer to database

Zac-HD:
Seamless handoff between symbolic and concrete tooling 🤩

### DIFF
--- a/crosshair/condition_parser.py
+++ b/crosshair/condition_parser.py
@@ -49,7 +49,7 @@ try:
     from hypothesis.internal.conjecture.data import ConjectureData
 except ModuleNotFoundError:
     hypothesis = None  # type: ignore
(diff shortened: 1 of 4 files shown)


pschanely:
One quick Q: is 6.0.0 the right minimum version given the shenanigans we're pulling? I bail out if hypothesis is requested, and the version is too old:
https://github.com/pschanely/CrossHair/blob/5f797f675bd161611baf33146a65dc7ed0e1aeae/crosshair/main.py#L83

Zac-HD:
Yep, 6.0 is a reasonable lower bound - the specific internals we're using go back a fair bit further, but you'd start to run into other interoperability problems and I'd rather not encourage that.  Hypothesis explicitly only supports our most recent release!

PR #335, opened 2025-02-28
allow os.chdir

KRRT7:
The bug was a logical contradiction in the code's permission system:

The code explicitly lists os.chdir as an allowed operation in one section
But later in the same function, it blocks all operations with the "os" prefix, which includes os.chdir

This meant that despite the intention to allow os.chdir (as indicated in the comments and the allowed events list), the prefix-based blocking was overriding that permission, causing os.chdir operations to be incorrectly blocked and raising the SideEffectDetectedError exception.


pschanely:
@KRRT7  As discussed in chat, I think this particular operation is currently allowed. I'd be happy to merge just the test though!

KRRT7:
I've updated to only include the test, I'll work on debugging the original problem and extracting the auditwall separately.

crosshair/auditwall.py
```diff
@@ -135,7 +135,6 @@ def make_handler(event: str) -> Callable[[str, Tuple], None]:
         "imaplib",
         "msvcrt",
         "nntplib",
-        "os",
```
pschanely, line 138: I think we should keep the `os` module blocked by default! Otherwise, this is good to go. Feel free to add yourself to the list in `contributing.rst` too!

pschanely (changes requested):

crosshair/auditwall.py
KRRT7: https://github.com/pschanely/CrossHair/blob/main/crosshair/auditwall.py#L131

https://github.com/pschanely/CrossHair/blob/main/crosshair/auditwall.py#L138

it's a duplicate hence the removal

pschanely (approved):
Thanks!

PR #337, opened 2025-03-13
Migrate test cases from unittest to pytest

### LINKED ISSUE #53: Migrate to pytest
Pytest seems like it might be better for us then unittest:
* it works better with mypy (because mypy understands real asserts)
* It's easier to selectively run tests (CrossHair has a lot of not-fast tests)

aksh16:
Pull request to resolve #53


aksh16:
Hi @pschanely , I see that `statespace_test.py::test_timeout` has failed but when I locally ran the tests, it did not fail.

How may I fix this? I did not directly change `statespace_test.py` but is it possible that this failure is because of the changes I pushed? For example, this test might be depending on one of the tests I updated from unittest to pytest.

pschanely:
It also passed in Python 3.8 through 3.12, so this is probably just a flake. (just checked on the exact 3.13.2 locally and passes for me to). I've triggered a github re-run just now and we'll see.

crosshair/core_test.py
```diff
@@ -1245,5 +1246,3 @@ def f(x: int) -> int:
         import cProfile
 
         cProfile.run("profile()", "out.pprof")
-    else:
-        unittest.main()
```
pschanely, line 1249: This one looks good to me, but in most other cases, we're ONLY running `unittest.main()`. In those cases, let's just remove the whole `if __name__ == "__main__":` block, because it won't do anything anyway!

aksh16: Thanks for the feedback! Should I not keep the `if __name__ == "__main__":` block where `set_debug(True)` is present in the main block? Because I kept the block thinking it might be serving some purpose

aksh16:
> Looks like it's still failing; I imagine we're just pushing over some sort of performance threshold on the github runners. (I bet it would fail in the same way on main if I triggered a run right now) I'll pursue this independently; don't worry about it for now.

I would be happy to look into it too. However, I might need access to the github runners and I am not sure how much compute the repo is allowed to consume to not cross the current tier. 
Still, if there is any way to help, I am up for it. I have worked with CI/CD jobs in the past and I am currently working on them too.

crosshair/core_test.py
pschanely: Ah. Yes that flag will turn on debugging, but the only way to enter that block is to run the file directly (`python crosshair/blah_test.py`), in which case it will just immediately exit anyway. Let's remove them!

pschanely:
> I would be happy to look into it too. However, I might need access to the github runners and I am not sure how much compute the repo is allowed to consume to not cross the current tier. Still, if there is any way to help, I am up for it. I have worked with CI/CD jobs in the past and I am currently working on them too.

Thanks; I just ran it again and it passed (on master and here). I assume it's just that the virtualization gets too slow under load, and exceeds my expected bounds. I've just up'd the bounds in [this commit](https://github.com/pschanely/CrossHair/commit/09b71cebf996cc18cf326e02e0553ca6b4e6367f). Ideally we might do something to manually manipulate the clock for the purposes of the test, but that's a thought for another day.

Don't forget to add yourself to to the contributor list in `contributing.md` if you like!

pschanely (approved):

aksh16:
Hi @pschanely , CI skipped upload_pypi, is that the reason the pull request did not get merged?

pschanely:
> Hi @pschanely , CI skipped upload_pypi, is that the reason the pull request did not get merged?

ah, no that's just skipped because it's not a release build. I just got distracted before clicking the merge button.

Thanks so much for this!

PR #409, opened 2026-05-12
Remove CompositeTracer wrapper layer (item 2 of #115)

islamicfinanceaaoifi-ops:
This PR addresses **item (2)** of [#115](https://github.com/pschanely/CrossHair/issues/115) — _"CompositeTracer is now largely just a wrapper around CTracer. Let's try to remove the CompositeTracer layer entirely."_

Item (3) (porting `TracingModule` to C) is left for a follow-up PR so the diff stays reviewable; happy to take that on next if this approach looks right.

## What changed

The old `CompositeTracer` class owned a `self.ctracer = CTracer()` instance and forwarded most methods through. The session lifecycle — push/pop the `patching_module`, wire `sys.monitoring` (3.12+) or `sys.settrace` (older), start/stop the tracer — lived in its `__enter__`/`__exit__`. This PR folds those responsibilities into `CTracer` itself and deletes the wrapper.

### C side (`_tracers.c`, `_tracers.h`)
- New `patching_module` slot on `CTracer`, exposed via `tp_getset`. Defaults to `None`.
- New `__enter__` / `__exit__` C methods that push the `patching_module`, configure `sys.monitoring` on 3.12+ (or stash `sys.gettrace()` and enable opcode tracing on the current frame for older Python), call `start()`/`stop()`, then unwind on exit.
- New `trace_caller` C method (no-op on 3.12+; on older Python enables opcode tracing on the calling frame — same as the previous `sys._getframe(2)`-based Python implementation).
- `CTracer.push_module` now invokes `sys.monitoring.restart_events()` itself on 3.12+ (only when the tracer is enabled — restarting has no effect otherwise). That `restart_events()` call used to live in the now-deleted `CompositeTracer.push_module`.
- Helpers `_ch_load_sys_monitoring` and `_ch_call_monitoring_method` centralize the `sys.monitoring` lookup so the lifecycle methods stay readable.

### Python side (`tracers.py`)
- `CompositeTracer` class is gone. The module singleton is now just `COMPOSITE_TRACER = CTracer()` with `patching_module = PatchingModule()` set explicitly.
- `is_tracing`, `NoTracing`, `ResumedTracing`, and `PushedModule` now go through the `CTracer` instance directly instead of `COMPOSITE_TRACER.ctracer`.

### Caller renames
The wrapper provided two name-only shims; callers move to the unified `CTracer` API:
- `COMPOSITE_TRACER.pop_config(m)` -> `COMPOSITE_TRACER.pop_module(m)`
- `COMPOSITE_TRACER.set_postop_callback(callback, frame)` -> `COMPOSITE_TRACER.push_postop_callback(frame, callback)` _(argument order swap)_

Touches `opcode_intercept.py` (7 callsites), `enforce.py`, `core.py`, `tracers_test.py`.

## Verified locally (Python 3.14.3 / Windows 11 / MSVC 14.50)

```
pytest crosshair/tracers_test.py crosshair/_tracers_test.py     # 12 / 12 pass (3 skipped baseline)
pytest crosshair/condition_parser_test.py                       # 27 / 27 pass
pytest crosshair/enforce_test.py crosshair/statespace_test.py   # 16 / 16 pass
pytest crosshair/util_test.py crosshair/opcode_intercept_test.py
pytest crosshair/fuzz_core_test.py                              # 1294 / 1294 pass
pytest crosshair/core_test.py crosshair/diff_behavior_test.py crosshair/path_cover_test.py   # 106 / 106 pass
```

`libimpl/` also runs through; the only failure I saw (`datetimelib_ch_test::test_builtin[check_datetime_astimezone]`) reproduces on unmodified `main` and isn't related to this refactor — current main CI is red for the same family of path-exploration failures.

## Design notes / questions for reviewer

- I kept the public name `COMPOSITE_TRACER` so this PR doesn't ripple a rename through downstream consumers. If you'd rather call it `TRACER` or similar now that it's just a `CTracer`, happy to follow up.
- `sys.monitoring.restart_events()` inside `push_module` is now guarded by `self->enabled`. The previous Python wrapper called it unconditionally. Let me know if you'd prefer the unconditional behavior — I went conservative on the perf side since it's only meaningful when the tracer is actively running.
- `_PyFrame_GetBackBorrow` in `trace_caller` walks back one frame to mirror the old `sys._getframe(2)` semantics (top Python frame on the eval stack is the caller of `trace_caller`, one back is the function we want to trace).
- `prev_traceobj` is stored on the struct for the older-Python branch so `__exit__` can restore `sys.settrace` to whatever was set before the session.

Ready for item (3) (porting `TracingModule.trace_op` + `_CALL_HANDLERS` and the `handle_call_*` family to C) as a follow-up — let me know if you'd like me to open that PR next.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

</details>

### DIFF
--- a/crosshair/_tracers.c
+++ b/crosshair/_tracers.c
@@ -129,6 +129,8 @@ CTracer_init(CTracer *self, PyObject *args_unused, PyObject *kwds_unused)
     init_framecbvec(&self->postop_callbacks, 5);
     init_modulevec(&self->modules, 5);
     init_tablevec(&self->handlers, 3);
+    self->patching_module = NULL;
+    self->prev_traceobj = NULL;
     self->enabled = FALSE;
     self->handling = FALSE;
     self->trace_all_opcodes = FALSE;
@@ -149,6 +151,8 @@ CTracer_dealloc(CTracer *self)
     PyMem_Free(self->postop_callbacks.items);
     PyMem_Free(self->modules.items);
     PyMem_Free(self->handlers.items);
+    Py_XDECREF(self->patching_module);
+    Py_XDECREF(self->prev_traceobj);
     Py_TYPE(self)->tp_free((PyObject*)self);
 }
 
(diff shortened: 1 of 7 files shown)


pschanely:
@islamicfinanceaaoifi-ops thanks for helping! Please check contributing.rst for getting the formatting/linting etc right.

> sys.monitoring.restart_events() inside push_module is now guarded by self->enabled

I expect that we'll still need this to happen because the tracer can be re-enabled later. If the tests pass anyway, I'd want to know whether we need test coverage there, or why my understanding is flawed! :)

pschanely:
Closing this due to inactivity post-review. But feel free to open a new one if you want to continue!

PR #410, opened 2026-05-13
Move call stack decoding into the C tracer

michael-schvarcz:
Addresses part of #115.

This is a focused slice of the remaining tracer hot-path work, separate from #409's `CompositeTracer` removal. It moves the call-opcode stack decoding currently done by Python `_CALL_HANDLERS` into the existing `_crosshair_tracers` C extension, then has `TracingModule.trace_op()` consume the new `call_stack_info()` helper.

The target normalization, kwargs symbolic-string handling, and replacement write-back behavior stay in Python for now, so this keeps the behavioral surface small while removing several Python function dispatches from every traced call opcode.

Tests run:

- `PYTHONHASHSEED=0 .venv/bin/python -m pytest crosshair/tracers_test.py crosshair/_tracers_test.py -q`
- `PYTHONHASHSEED=0 .venv/bin/python -m pytest crosshair/core_test.py crosshair/libimpl/builtinslib_test.py crosshair/tracers_test.py crosshair/_tracers_test.py -q`
- `.venv/bin/python -m black --check crosshair/tracers.py`
- `.venv/bin/python -m isort --check-only crosshair/tracers.py`

AI-assisted with Codex, with local build and tests run before submission.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#115: Improve trace callback performance](https://oss.issuehunt.io/repos/101762821/issues/115)
---
</details>

PR #411, opened 2026-05-14
Move call target normalization into C tracer

iamdinhthuan:
This is a focused follow-up slice for #115 after #410 moved call stack decoding into the C extension.

`TracingModule.trace_op()` still normalized bound methods and callable instances in Python on every traced call opcode. This moves that target-normalization step into `_crosshair_tracers.normalize_call_target()`, while keeping the existing Python `trace_call()` override hook, keyword handling, and stack write-back behavior unchanged.

What changed:
- Added a C helper that mirrors the existing `object.__getattribute__`-based normalization of bound methods and callable instances.
- Replaced the Python normalization block in `TracingModule.trace_op()` with the C helper.
- Added coverage for bound methods, callable instances, and refcount stability.

Verification:
- `PYTHONHASHSEED=0 python -m pytest crosshair/tracers_test.py crosshair/_tracers_test.py -q` -> `13 passed, 2 skipped`
- `PYTHONHASHSEED=0 python -m pytest crosshair/core_test.py crosshair/libimpl/builtinslib_test.py crosshair/tracers_test.py crosshair/_tracers_test.py -q` -> `556 passed, 6 skipped`
- `python -m black --check crosshair/tracers.py crosshair/tracers_test.py`
- `python -m isort --check-only crosshair/tracers.py crosshair/tracers_test.py`
- `python -m flake8 crosshair/tracers.py crosshair/tracers_test.py`
- `git diff --check`

This intentionally avoids touching the open CompositeTracer-removal work in #409.

### DIFF
--- a/crosshair/_tracers.c
+++ b/crosshair/_tracers.c
@@ -1262,6 +1262,142 @@ static PyObject *crosshair_tracers_call_stack_info(PyObject *self, PyObject *arg
     Py_RETURN_NONE;
 }
 
+static PyObject *
+crosshair_tracers_getattr_or_none(PyObject *obj, const char *name)
+{
+    PyObject *name_obj = PyUnicode_InternFromString(name);
+    if (name_obj == NULL) {
+        return NULL;
+    }
+    PyObject *ret = PyObject_GenericGetAttr(obj, name_obj);
+    Py_DECREF(name_obj);
+    if (ret == NULL && PyErr_ExceptionMatches(PyExc_AttributeError)) {
+        PyErr_Clear();
+        Py_RETURN_NONE;
+    }
+    return ret;
+}
+
+static PyObject *
+crosshair_tracers_normalize_call_target(PyObject *self, PyObject *args)
+{
+    PyObject *target;
+    PyObject *selfless_callable_types;
+    PyObject *normal_callable_types;
+    if (!PyArg_ParseTuple(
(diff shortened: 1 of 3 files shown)


crosshair/_tracers.c
```diff
@@ -1262,6 +1262,142 @@ static PyObject *crosshair_tracers_call_stack_info(PyObject *self, PyObject *arg
...
+                Py_XDECREF(owned_target);
+                return NULL;
+            }
+            if (target_name != Py_None) {
+                PyObject *typelevel_target = PyObject_GetAttr(
+                    (PyObject *)Py_TYPE(bound_self), target_name);
+                if (typelevel_target == NULL &&
+                        PyErr_ExceptionMatches(PyExc_AttributeError)) {
+                    PyErr_Clear();
+                }
```
pschanely, line 1372: Unlikely, but could cause crashes/corruption: After this check we should check for a non-AttributeError exception (rare, but could happen from complex descriptor access logic). Just propagate that kind of error with something like:
```
if (typelevel_target == NULL) {
    Py_DECREF(target_name); Py_DECREF(func);
    Py_DECREF(bound_self); Py_XDECREF(owned_target);
    return NULL;
}
```

pschanely (changes requested):
Very exciting; just one change request below.

PR #416, opened 2026-05-24
Support for arbitraty generics

### LINKED ISSUE #415: Support generics in function signature
**Expected vs actual behavior**
It seems that `crosshair` can't handle generics outside very specific things like `list`.

I don't know if it's a fundamental limitation or if it's just a simple bug to fix. In any case the error is not very helpful, and there should be at least a way to silence it.

**To Reproduce**
```py
"""Minimal reproduction of crosshair error with generic types."""

from typing import Generic, TypeVar

T = TypeVar("T")


class Container(Generic[T]):
    # meant to illustrate something like torch.utils.data.DataLoader
    pass


def process(container: Container[int]) -> int:
    """Process container.

    pre: container is not None
    """
    pass
```

rambip:
This is my attempt to fix #415 

To be clear:
- I was in a rush, the solution was LLM-generated
- I did not review the code
- I only tested on my toy-example

If you consider it's slop and close it, I will have no problem with that.

Have a nice day !

### DIFF
--- a/crosshair/core.py
+++ b/crosshair/core.py
@@ -515,8 +515,18 @@ def proxy_for_class(typ: Type, varname: str) -> object:
         raise CrosshairUnsupported(
             f"unable to create concrete instance of {typ} due to bad constructor"
         )
-    # TODO: use dynamic_typing.get_bindings_from_type_arguments(typ) to instantiate
-    # type variables in `constructor_sig`
+    bindings = dynamic_typing.get_bindings_from_type_arguments(typ)
+    if bindings:
+        new_params = []
+        for p in constructor_sig.parameters.values():
+            if p.annotation != inspect.Parameter.empty:
+                try:
+                    resolved = dynamic_typing.realize(p.annotation, bindings)
+                    p = p.replace(annotation=resolved)
+                except KeyError:
+                    pass  # unbound TypeVar — leave annotation as-is
+            new_params.append(p)
+        constructor_sig = constructor_sig.replace(parameters=new_params)
     args = gen_args(constructor_sig)
     typename = name_of_type(typ)
     try:
--- a/crosshair/core.py
+++ b/crosshair/core.py
@@ -499,7 +499,7 @@ def proxy_for_class(typ: Type, varname: str) -> object:
     if data_members is None:
         try:
             data_members = get_type_hints(cls)
-        except Exception:
+        except (AttributeError, NameError):
             # Forward references that can't be resolved outside the defining module
             # (e.g. torch.utils.data.DataLoader) cause NameError/AttributeError here.
             data_members = {}
@@ -525,11 +525,8 @@ def proxy_for_class(typ: Type, varname: str) -> object:
         new_params = []
         for p in constructor_sig.parameters.values():
             if p.annotation != inspect.Parameter.empty:
-                try:
-                    resolved = dynamic_typing.realize(p.annotation, bindings)
-                    p = p.replace(annotation=resolved)
-                except KeyError:
-                    pass  # unbound TypeVar — leave annotation as-is
+                resolved = dynamic_typing.realize(p.annotation, bindings)
+                p = p.replace(annotation=resolved)
             new_params.append(p)
         constructor_sig = constructor_sig.replace(parameters=new_params)
     args = gen_args(constructor_sig)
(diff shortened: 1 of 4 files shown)


rambip:
~Hum, it does not seem to solve this case right now:~
 now it does
```
def accuracy(model: nn.Module, loader: DataLoader[tuple[Tensor, Tensor]]) -> float:
    """Compute accuracy on a dataset.

    pre: loader is not None and len(loader) > 0
    inv: 0 <= return_value <= 1
    """
    correct = 0
    total = 0
    for X, y in loader:
        preds = model(X).argmax(dim=1)
        correct += (preds == y).sum().item()
        total += len(y)
    return correct / total
```

crosshair/core.py
```diff
@@ -490,33 +490,53 @@ def get_constructor_signature(cls: Type) -> Optional[inspect.Signature]:
...
+        cls = typ
+    data_members = _TYPE_HINTS.get(cls, None)
     if data_members is None:
-        data_members = get_type_hints(typ)
-        _TYPE_HINTS[typ] = data_members
-
-    if sys.version_info >= (3, 8) and type(typ) is typing._TypedDictMeta:  # type: ignore
+        try:
+            data_members = get_type_hints(cls)
+        except Exception:
```
pschanely: Let's just catch the ones we think are likely.
```suggestion
        except (AttributeError, NameError):
```

crosshair/core.py
```diff
@@ -490,33 +490,53 @@ def get_constructor_signature(cls: Type) -> Optional[inspect.Signature]:
...
 
-    constructor_sig = get_constructor_signature(typ)
+    constructor_sig = get_constructor_signature(cls)
     if constructor_sig is None:
         raise CrosshairUnsupported(
             f"unable to create concrete instance of {typ} due to bad constructor"
         )
-    # TODO: use dynamic_typing.get_bindings_from_type_arguments(typ) to instantiate
-    # type variables in `constructor_sig`
+    bindings = dynamic_typing.get_bindings_from_type_arguments(typ)
```
pschanely, line 523: [note] `get_bindings_from_type_arguments` is unused until now, and I'm fairly confident it's not fully correct. But that shouldn't stop us from handling at least some cases, so I'm ok with this.

crosshair/core.py
```diff
@@ -490,33 +490,53 @@ def get_constructor_signature(cls: Type) -> Optional[inspect.Signature]:
...
+    if bindings:
+        new_params = []
+        for p in constructor_sig.parameters.values():
+            if p.annotation != inspect.Parameter.empty:
+                try:
+                    resolved = dynamic_typing.realize(p.annotation, bindings)
+                    p = p.replace(annotation=resolved)
+                except KeyError:
```
pschanely: It's probably better to push this fallback down into `dynamic_typing.resolve` rather than catch KeyError. Then we can resolve some vars inside a parameterized type even if one of them isn't bound. `resolve()` ALSO isn't used anywhere presently, so we don't need to worry about changing behavior.

crosshair/core.py
```diff
@@ -490,33 +490,53 @@ def get_constructor_signature(cls: Type) -> Optional[inspect.Signature]:
 
 
 def proxy_for_class(typ: Type, varname: str) -> object:
```
pschanely, line 492: Please add tests to cover the new changes. (it's likely fine to have Claude do it!)

pschanely (changes requested):
Not slop, as far as I can tell. Thank you for the contribution!
I have a few (hopefully small) requests; feel free to add yourself at the bottom of `contributing.rst` too.

rambip:
Is that ok now ?

pschanely:
> Is that ok now ?

Just a few more mypy typing issues, but otherwise this looks great!
FWIW, in many cases, I just remove the `-> None` from the test definition to disable checking for tests that are painful to get compliant.

rambip:
I'm confused about the `mypy` errors, some seems unrelated to my change.

pschanely:
> I'm confused about the `mypy` errors, some seems unrelated to my change.

Unfortunately there are a lot of warning messages in there (I'll file an issue about this presently); for now, you only need to look at the errors.

rambip:
Only hypothesis failing

pschanely (approved):
LGTM! Thank you!

PR #425, opened 2026-06-01
Keep date/datetime/time/timedelta comparisons symbolic

### LINKED ISSUE #424: Ensure `datetime.date` comparisons remain symbolic
When CrossHair is used as a Hypothesis backend (`hypothesis-crosshair`), symbolic `datetime.date` values from the stock `st.dates()` strategy get concretized during generation, so the backend degrades to random search over dates. The root cause is `calendar.monthrange`, which Hypothesis calls to bound the day-of-month draw: it does `mdays[month]`, indexing a concrete list with the symbolic month, which realizes `month`. CrossHair already avoids this exact trap internally (`datetimelib._days_in_month` is commented *"Avoid `_DAYS_IN_MONTH` so that we don't realize the month"*), but `calendar` isn't patched, so it sneaks back in via the strategy. A small `libimpl/calendarlib.py` fixes it with no change to Hypothesis.

Alternatively, https://github.com/pschanely/CrossHair/issues/332 would at least help and might fully solve the issue.

## Reproduction

```python
from datetime import date
from hypothesis import given, settings, strategies as st

@settings(backend="crosshair", deadline=None, database=None)
@given(st.dates())
def test_date_needle(d):
    assert d != date(2030, 2, 14)  # PASSES — counterexample MISSED
```

## Diagnosis

`DateStrategy.do_draw` → `draw_capped_multipart` bounds the day draw with `_, high = monthrange(year, month)` where `year`/`month` are symbolic.

`calendar.monthrange` computes `ndays = mdays[month] + (month == FEBRUARY and isleap(year))`; the concrete-list index `mdays[month]` forces realization of `month`. Once `month` is realized the solver can't choose it to satisfy a downstream constraint like `d.month == 2`, so date search collapses to random-like. (It also builds a `datetime.date` via `weekday(...)` that the caller discards.)  Everything else in the path is already symbolic-safe.

## Proposed fix

Add `crosshair/libimpl/calendarlib.py` registering symbolic-friendly replacements for `monthrange` (and `_monthlen`/`weekday`), mirroring `datetimelib._days_in_month`. Proof of concept:

```python
from crosshair import register_patch
import calendar

def sym_monthrange(year, month):
    if month == 2:
        leap = (year % 4 == 0) and (year % 100 != 0 or year % 400 == 0)
        ndays = 29 if leap else 28
    elif month == 4 or month == 6 or month == 9 or month == 11:
        ndays = 30
    else:
        ndays = 31
    return 0, ndays   # PoC: dummy weekday; a real patch computes it symbolically

register_patch(calendar.monthrange, sym_monthrange)
```

With only this registered, the **unchanged** `st.dates()` needle now FAILS in ~6s, with CrossHair pinning `year=2030, month=2, day=14` — the full date is symbolic again.

## Motivation

I recently found [_DateSAT: A Framework for Solving Date and Period Constraints_](https://arxiv.org/pdf/2605.25180), and in the course of checking that `hypothesis-crosshair` did not trivially solve their benchmark I discovered that we have this earlier problem!  

And then also don't solve their benchmark, so we might want to consider adopting the techniques in their paper to reason about dates in crosshair, but that's another issue.  (and also work out how to have hypothesis-crosshair say "these bits represent a date, these a timedelta" etc, and maybe handle datetimes too... _ars longa, vita brevis_)

Zac-HD:
The comparison dunders delegated to a module-level _cmp that returned a concrete -1/0/1 by comparing tuples with Python's tuple </==.  That tuple comparison forked on each element to coerce SymbolicBool back to a real bool, so the solver saw a chain of sequential decisions rather than a single constraint.  Separately, none of these classes defined __ne__, so `!=` fell back to Python negating __eq__ via bool(), realizing the operand too.

Make _cmp return a symbolic three-way value, built from a lexicographic less-than (_lex_lt) over the components using the patched any/all builtins (which collapse to single SMT terms).  Every `self._cmp(other) <op> 0` then stays symbolic, with no per-operator changes; timedelta (which routes through the same _cmp) benefits too.  Add __ne__ methods (date, datetime, time, timedelta, timezone) that negate the symbolic __eq__ result with smt_not instead of bool().

Closes https://github.com/pschanely/CrossHair/issues/424; the remaining problems are an issue for how Hypothesis constructs these objects.

### DIFF
--- a/crosshair/libimpl/datetimelib.py
+++ b/crosshair/libimpl/datetimelib.py
@@ -35,13 +35,34 @@ from crosshair import (
     register_type,
 )
 from crosshair.core import SymbolicFactory
-from crosshair.libimpl.builtinslib import make_bounded_int, smt_or
+from crosshair.libimpl.builtinslib import make_bounded_int, smt_not, smt_or
 from crosshair.statespace import context_statespace
 from crosshair.tracers import NoTracing
 
 
+def _lex_lt(left, right):
+    """Lexicographic strict-less-than over two equal-length sequences."""
+    clauses = []
+    for i in range(len(left)):
+        clause = [left[j] == right[j] for j in range(i)]
+        clause.append(left[i] < right[i])
+        clauses.append(all(clause))
+    return any(clauses)
+
+
 def _cmp(x, y):
-    return 0 if x == y else 1 if x > y else -1
+    # Three-way comparison built from `any`/`all` (which the patched builtins
+    # collapse to single SMT terms), so comparisons of symbolic dates/times
+    # stay symbolic instead of realizing each component via tuple comparison.
+    return _lex_lt(y, x) - _lex_lt(x, y)
+
+
+def _symbolic_ne(eq_result):
+    # Negate an __eq__ result while keeping it symbolic.  (Python's default
+    # __ne__ calls bool() on the __eq__ result, which would realize it.)
+    if eq_result is NotImplemented:
+        return NotImplemented
+    return smt_not(eq_result)
 
 
 MINYEAR = 1
@@ -754,6 +775,9 @@ class timedelta:
         else:
             return NotImplemented
 
+    def __ne__(self, other):
+        return _symbolic_ne(self.__eq__(other))
+
     def __le__(self, other):
         if isinstance(other, any_timedelta):
             return self._cmp(other) <= 0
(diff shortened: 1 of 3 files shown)


pschanely:
This is so great. Impressed that claude (or you) figured out the `any`/`all` trick!

Zac-HD:
that was Claude!  _several_ false starts, cf #424 being about `calendar`, but Opus 4.8 is pretty good and we got there in the end 😁

PR #446, opened 2026-07-02
Fix Python 3.15 CI failures (differential suite, base64, decimal)

pschanely:
Follow-up to #434, addressing the 10 failures surfaced by the 3.15-dev CI job. All 10 are resolved; verified on a local 3.15.0b3 build (both with and without the C `_decimal` extension).

## Commits

- **Classify two new-in-3.15 builtins** — `bytearray.take_bytes` (xfail, like the 3.14 `bytearray.resize` gap) and `builtins.__lazy_import__` (differential-skip, like `__import__`). The differential fuzzer enumerates every builtin/method, so both newcomers started failing.

- **Support 3.15's new `binascii.b2a_base64` keyword arguments** — 3.15 gave `b2a_base64` `wrapcol`/`padded`/`alphabet` and rerouted `base64.encodebytes`/`b64encode`/`urlsafe_b64encode` through them, so every base64 *encode* path raised `TypeError: unexpected keyword argument 'wrapcol'`. The patch now implements all three (the `wrapcol` newline algorithm and `alphabet` remap were checked exhaustively against the real interpreter), keeping the mapping symbolic-friendly.

- **`devcontainer`: install `libmpdec-dev`** — Python 3.15 unbundled libmpdec, so the C `_decimal` extension only builds against a system copy. Without it, `decimal` silently falls back to pure-Python `_pydecimal`. Installing the dev package keeps built interpreters on C `_decimal`, matching properly-provisioned CPython.

- **Don't model `decimal` when the pure-Python `_pydecimal` is active** — CrossHair's `Decimal` shim mirrors the C `_decimal` extension, so on libmpdec-less builds it diverged from the runtime actually in use (strict-vs-lenient constructor validation, `int()` on a symbolic coefficient). Rather than model it unsoundly, `make_registrations` now registers no decimal patches in that case: concrete decimal code still runs, symbolic `Decimal` args become unsupported. This also drops the `patch_equivalence[Decimal-*]` params (they iterate `_PATCH_REGISTRATIONS`) and skips the two decimal test modules on such builds.

- **`devcontainer`: add the GitHub CLI feature** — installs `gh` in the container (unrelated to 3.15, bundled here for convenience).

## Verification

- **C `_decimal` present** (3.13, and a 3.15.0b3 rebuilt with `libmpdec-dev`): full decimal support runs and passes; no regressions.
- **`_decimal` absent** (CI's 3.15): decimal patches drop, `patch_equivalence` collects 0 Decimal params, decimal test modules skip; base64 and builtin classifications hold.

Note: CI's 3.15 job is `continue-on-error`, and its hosted-toolcache interpreter currently lacks `_decimal`, so the decline-to-model path is what keeps it green there.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01H9V78EaWcUtDKJF7Z6MNN4

### DIFF
--- a/crosshair/libimpl/decimallib.py
+++ b/crosshair/libimpl/decimallib.py
@@ -5246,6 +5246,24 @@ def make_function_with_mapped_args(fn):
 
 
 def make_registrations():
+    import _pydecimal
+
+    # Our Decimal shim reimplements _pydecimal in a symbolic-friendly way, but is
+    # written and validated against the C `_decimal` extension.  When the
+    # interpreter instead falls back to the pure-Python `_pydecimal` (builds
+    # without a system libmpdec -- e.g. Python 3.15, which unbundled it), the
+    # shim's *concrete* behavior no longer matches the `decimal` actually
+    # running: e.g. our constructor validates its context argument eagerly, like
+    # C `_decimal`, whereas `_pydecimal` does not.  Reasoning about a program
+    # with a model that disagrees with its real runtime is unsound, so we
+    # register nothing.  `decimal` then runs unpatched: concrete Decimal values
+    # behave exactly like the real module, and a *symbolic* Decimal argument
+    # degrades to CrosshairUnsupported (we can't build a symbolic proxy without
+    # the registration), so that path is skipped rather than mismodeled.
+    if real_decimal.Decimal is _pydecimal.Decimal:
+        debug("Pure-Python _pydecimal is active; not modeling decimal (see comment)")
+        return
+
     # "DecimalTuple",  # do I want this?
     register_patch(real_decimal.Decimal, lambda *a, **kw: Decimal(*a, **kw))
     register_type(real_decimal.Decimal, _make_decimal)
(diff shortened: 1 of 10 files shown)


crosshair/libimpl/decimallib.py
```diff
@@ -5246,6 +5246,21 @@ def wrapper(self_context, *a, **kw):
...
+
+    # CrossHair's Decimal shim mirrors the behavior of the C `_decimal`
+    # extension.  When the interpreter instead falls back to the pure-Python
+    # `_pydecimal` -- which happens on builds without a system libmpdec, e.g.
+    # Python 3.15 unbundled it -- that shim can diverge from the decimal
+    # implementation actually running, which would make symbolic decimal
+    # reasoning unsound.  So decline to model decimal at all in that case:
+    # concrete decimal code still runs correctly (unpatched) against the real
+    # `_pydecimal`, and functions with symbolic Decimal arguments simply become
+    # unsupported rather than silently wrong.
```
pschanely: Not sure if I follow this exactly. To be pedantic, values can be concrete, but "concrete code" isn't really a thing. Furthermore, even using _pydecimal directly, functions with symbolic arguments may still work just fine, because (in theory) the rest of Python is supported. That said, we customize our decimal implementation to be symbolic-friendly, so I might expect degradation in the fallback case, and unit tests may fail that expect certain symbolic behaviors. (OOC, do you know what the test results looked like for this case?)

Issue #47, opened 2021-01-27
Allow for the possibility for aliased top-level parameter values


pschanely:
One of CrossHair's strengths is that it models the heap and can detect a variety of "aliasing" problems - bugs caused by having the same object accessible through different paths of references. [In this example](https://crosshair-web.org/?crosshair=0.1&python=3.8&gist=130e7ff2373652556718f1df5ce57f4d), we can detect that two sub-lists might be the same list, which breaks the expected postcondition.

But CrossHair only does this for container contents, not for top-level parameters. For instance, it does not understand that the two parameters of [this function](https://crosshair-web.org/?crosshair=0.1&python=3.8&gist=17a569c4832e2a8244c7bbda40800a5f) could be the same list.

pschanely:
The operation-differential test suite now actively **witnesses** this gap. A custom aliased-input strategy feeds `(x, x)` — the *same* object in both positions — to `operator.is_` / `operator.is_not`, and the symbolic-vs-concrete differential diverges:

```
operator.is_     on ([0], [0]):  concrete → True,   symbolic → False
operator.is_not  on ([0], [0]):  concrete → False,  symbolic → True
```

i.e. symbolic execution treats two separate top-level parameters as necessarily-distinct objects, exactly as described here. These are currently recorded as known-failing (xfail) in the differential test.

Note there is a coupled change on the *test* side when this is fixed: the differential harness pins each parameter to its own symbolic proxy, so even once the engine models top-level aliasing, the harness must pin an aliased concrete input to a **shared** proxy for these xfails to flip to passing.

PR #457, opened 2026-07-07
Unify operation classification into one auditable catalog

pschanely:
Unifies the operation-classification layer that the differential fuzz test and the support-map measurement had been duplicating (with separate, drifting module lists and ad-hoc decisions).

## What changed

**One catalog** (`crosshair.inputgen.catalog()`) yields an `Operation` per op with five orthogonal classification fields — `skip_reason`, `out_of_scope`, `not_value_function`, `side_effect`, `probe_hazard`. Both `fuzz_core_test` and `measure_support` now enumerate it, so their coverage can’t drift.

- **Retire `behavior_compare.DIFFERENTIAL_SKIP`**: split by concept into `inputgen.NOT_VALUE_FUNCTION` (order/identity/reflection) and `SIDE_EFFECT_OVERRIDES` (builtin I/O), read off each `Operation`. Making `open`/`print` static side-effects means the sweep short-circuits them instead of fuzzing them forward (no more junk files).
- **Relocate** `measure_support._deterministic` → `inputgen.is_deterministic` (the dynamic half of `not_value_function`).
- **measure_support**: collapse the `surface`/`funcs`/`methods` commands (each with its own module list) into one `measure` command over the catalog, with `--tiers`/`--modules`/`--types` scoping. Keep the forkserver kill-harness.
- **auditwall**: add `enabled_auditwall()` so the concrete side-effect probe can police a short region without leaving the wall active.

**`CUSTOM_INPUTS` registry** (generalizes the former `_ROUNDTRIP`): seedkey-keyed whole-tuple input strategies for ops where independent per-arg fuzzing can’t build a useful input, consulted by *both* input paths so a custom input is tested *and* performance-measured.

- `getattr(o, name)`: draw `name` from `o`’s real attributes; it’s now a passing CI-gate test (removed from `NOT_VALUE_FUNCTION`).
- `operator.is_`/`is_not`: draw an aliased `(x, x)`, which witnesses [#47](https://github.com/pschanely/CrossHair/issues/47) (top-level parameter aliasing unmodeled). Recorded as `xfail`; the coupled harness change (shared-proxy pinning) is noted there.

## Testing

CI fuzz gate: **390 passed, 1466 skipped, 8 xfailed, 4 xpassed**. Support map unchanged except `getattr`/`is_`/`is_not` now render as measured cells and `open`/`print` as explained side-effect cells.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crosshair/behavior_compare.py
+++ b/crosshair/behavior_compare.py
@@ -352,12 +352,6 @@ _UNSUPPORTED = object()  # pinned fine, but the op rejected the symbolic proxy
 _DIFF_MAX_PIN_ITERS = 80
 _DIFF_PIN_TIMEOUT = 10.0
 
-# (Operations the differential can't meaningfully check -- unordered-container
-# ordering, identity-eq / non-value-comparable output, reflection -- were listed
-# here as DIFFERENTIAL_SKIP.  They now live in the operation catalog's
-# ``not_value_function`` / ``side_effect`` classification: crosshair.inputgen's
-# NOT_VALUE_FUNCTION and SIDE_EFFECT_OVERRIDES, read off each ``Operation``.)
-
 
 @dataclass
 class Divergence:
--- a/crosshair/behavior_compare.py
+++ b/crosshair/behavior_compare.py
@@ -488,13 +488,16 @@ def run_differential(
     k: int = 3,
     seed: int = 0,
     max_pin_iters: int = _DIFF_MAX_PIN_ITERS,
+    seedkey: Optional[str] = None,
 ) -> DiffResult:
     """Drive one operation on up to ``k`` valid inputs, comparing a symbolic run
     (args pinned to the input) against a concrete run.  Returns a ``DiffResult``
     with the count of inputs actually driven and the first ``Divergence`` (or
     None if all matched).  ``max_pin_iters`` bounds the per-input pin search; use
     a small value when speed matters more than pinning every container shape (an
-    input that can't pin in budget is simply skipped).
+    input that can't pin in budget is simply skipped).  ``seedkey`` is the op's
+    catalog identity, forwarded to ``valid_inputs`` so a CUSTOM_INPUTS override
+    (e.g. aliased ``x is x``) applies.
 
     ``expr`` is eval'd over ``arg_names`` (plus ``eval_globals``); see
     ``crosshair.inputgen.op_call``/``func_call`` for building these."""
@@ -503,7 +506,11 @@ def run_differential(
     def applier(*vs):
         return eval(expr, dict(eval_globals), dict(zip(arg_names, vs)))
 
-    inputs = [t for t in valid_inputs(fn, k=k, seed=seed) if len(t) == len(arg_names)]
+    inputs = [
+        t
+        for t in valid_inputs(fn, k=k, seed=seed, seedkey=seedkey)
+        if len(t) == len(arg_names)
+    ]
     checked = 0
     unsupported = 0
     for vals in inputs:
--- a/crosshair/behavior_compare.py
+++ b/crosshair/behavior_compare.py
@@ -352,45 +352,11 @@ _UNSUPPORTED = object()  # pinned fine, but the op rejected the symbolic proxy
 _DIFF_MAX_PIN_ITERS = 80
 _DIFF_PIN_TIMEOUT = 10.0
 
-# Operations the differential CANNOT meaningfully check: their forward output
-# legitimately depends on unspecified iteration order, isn't value-comparable, or
-# the builtin isn't a pure value transform.  A divergence here is NOT a bug -- so
-# both the fuzz test (skips them) and the support measurement (won't color them
-# black from the differential) consult this list.  Keyed by "<type>.<method>" /
-# "<module>.<func>".
-DIFFERENTIAL_SKIP: Dict[str, str] = {
-    # repr/str of an unordered container -- element order in the string is arbitrary
-    "set.__repr__": "set repr element order is unspecified",
-    "frozenset.__repr__": "frozenset repr element order is unspecified",
-    "dict.__repr__": "dict repr order need not match between symbolic and concrete",
-    "set.__str__": "set str element order is unspecified",
-    "frozenset.__str__": "frozenset str element order is unspecified",
-    "dict.__str__": "dict str order need not match between symbolic and concrete",
-    # pop an arbitrary element -- which one is unspecified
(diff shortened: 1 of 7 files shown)


crosshair/inputgen.py
```diff
@@ -1258,3 +1364,538 @@ def func_call(
...
+    def inputs(self, k: int = 5, seed: int = 0, size: int = 3) -> List[Tuple[Any, ...]]:
+        """Concrete valid argument tuples for this op at the given ``size``."""
+        if self.call is None:
+            return []
+        return valid_inputs(
+            self.call[0], k=k, seed=seed, size=size, seedkey=self.seedkey
+        )
+
+
+def _static_skip_reason(
```
pschanely: Consider inlining.

crosshair/inputgen.py
```diff
@@ -1050,27 +1068,96 @@ def _hash_names() -> Tuple[str, ...]:
...
+#     base data) -- the former _ROUNDTRIP, now :func:`_roundtrip` entries;
+#   * ``getattr(o, name)`` wants ``name`` to be an attribute OF ``o`` -- a random
+#     string is almost always an AttributeError;
+#   * ``x is x`` / ``x is not x`` needs the SAME object in both positions, which
+#     independent draws never produce.
+# Each value is ``(specs, size) -> strategy-over-the-full-arg-tuple``; it may build
+# on the op's own per-arg ``specs`` (so it adapts to whatever arity a caller drives
+# the op with) or ignore them.  BOTH input paths consult this -- valid_inputs (the
+# differential) and the support sweep -- so a custom input is both tested and
+# performance-measured, not merely displayed.
```
pschanely: Documentation nit: please strike final sentence - describe everything based on what it does, independent from how it's used.

crosshair/inputgen.py
```diff
@@ -1258,3 +1364,538 @@ def func_call(
...
+
+    ``probe`` selects the LIVE side-effect probe per drivable op (static
+    classification -- skip / out_of_scope / hardcoded hazard -- always applies):
+      * ``False``      -- no live probe (the safe default; fast, and complete on a
+                          pure surface where nothing new is discovered).
+      * ``True``       -- in-process (fast; safe on a vetted surface, but a
+                          non-overridden blocking op will wedge the caller).
+      * ``"isolated"`` -- each op in a killable subprocess (safe on ANY surface;
+                          a blocking op is tagged :data:`HANG`).  Slower; forks,
+                          so run it from a clean process."""
```
pschanely, line 1856: Should we even support `probe`? It isn't actively used. I think your idea is that we would use this to maintain our "probe hazard" lists? Should we have a probing unit test to ensure our hazard list is accurate?

crosshair/inputgen.py
```diff
@@ -1258,3 +1364,538 @@ def func_call(
...
+        skip_reason=skip,
+        out_of_scope=oos,
+        not_value_function=nvf,
+        side_effect=side_effect,
+        probe_hazard=hazard,
+    )
+
+
+# The canonical operation surface -- the ONE definition both the support map and
+# the differential fuzz test enumerate, so their coverage can't drift.  Free
```
pschanely: Also strike historically-motivated docs/comments: keep justification & litigation in PRs, not code.

crosshair/behavior_compare.py
```diff
@@ -352,45 +352,11 @@ def compare_results(fn: Callable, *a: object, **kw: object) -> ResultComparison:
...
-    "builtins.globals": "introspects the caller's namespace",
-    "builtins.locals": "introspects the caller's namespace",
-    "builtins.vars": "introspects an object's namespace",
-    "builtins.dir": "introspection; order/content not a pure value",
-}
+# (Operations the differential can't meaningfully check -- unordered-container
+# ordering, identity-eq / non-value-comparable output, reflection -- were listed
+# here as DIFFERENTIAL_SKIP.  They now live in the operation catalog's
+# ``not_value_function`` / ``side_effect`` classification: crosshair.inputgen's
+# NOT_VALUE_FUNCTION and SIDE_EFFECT_OVERRIDES, read off each ``Operation``.)
```
pschanely: Here also - no need to justify decisions in code.

crosshair/inputgen.py
```diff
@@ -1258,3 +1364,538 @@ def func_call(
...
+            continue
+        out.append(name)
+    return out
+
+
+# ---------------------------------------------------------------------------
+# the operation catalog: ONE uniform record per operation, so the support
+# measurement and the differential fuzz test cover the *same* set instead of
+# three hand-drifting module lists.  Classification decisions are moved FORWARD
+# into cataloging here, as five ORTHOGONAL facts per op:
```
pschanely: "orthogonal" got me confused for a moment - do we expect it's ok to set more than one reason string? I assume no.

crosshair/inputgen.py
```diff
@@ -1258,3 +1364,538 @@ def func_call(
...
+#     BEFORE the effect, so we learn "this reaches for I/O" WITHOUT performing it.
+#     A neutral FACT, not a verdict: a future symbolic read()/urandom() (cf.
+#     symbolic time.time()) could make such an op supported.  The builtin I/O ops
+#     on the always-measured surface are hardcoded in SIDE_EFFECT_OVERRIDES so
+#     ``probe=False`` classifies them without a live probe.
+#   * probe_hazard -- the concrete probe can't be run: the op blocks (HANG) or
+#     crashes the interpreter (CRASH), so its nature is hardcoded, not observed.
+# The *dynamic* "unsupported" verdict (proxy intolerance / can't pin) still
+# belongs to the measurement pass -- it needs symbolic execution, which inputgen
+# neither does nor depends on.
```
pschanely: Strike. No need to talk about things this module doesn't depend on.

crosshair/inputgen.py
```diff
@@ -1258,3 +1364,538 @@ def func_call(
...
+@dataclass
+class Operation:
+    """One catalogued operation, uniform across builtin methods, stdlib class
+    methods, and module-level free functions.
+
+    ``call`` is the drive spec every consumer shares -- ``(fn, expr, arg_names,
+    eval_globals)`` -- or None when the op isn't drivable.  ``seedkey`` is the
+    dotted identity (``"<owner>.<name>"``) used for the fuzz seed and static-set
+    lookups; ``key`` is the rendered form the support map / usage prior use.  The
+    five classification fields are orthogonal (see the module comment above)."""
```
pschanely: Probably fold the giant comment block into our docstring?

PR #482, opened 2026-07-18
Faster symbolic indexing into some concrete sequences

Zac-HD:
While looking at follow-ups to #470, I stumbled across an upstream issue and opened https://github.com/Z3Prover/z3/issues/10162.

The remain patch is an additional test case, and some speedups to skip consideration of negative indices.


pschanely:
Nice! I wanted to do it just a little differently, so dropped a few extra commits on top. Local measurement says that performance improvement is preserved.

PR #484, opened 2026-07-18
Subscript, contain-check, and index ranges symbolically

Zac-HD:
Subscripting a concrete range with a symbolic int previously realized the index, pinning the element to a concrete value and forcing a solver query per subscript. `SymbolicRange.__getitem__` likewise realized the whole range for non-slice indexes.

This is a hot path for hypothesis-crosshair users via `st.sampled_from(range(...))`: with `sets(sampled_from(range(n)), min_size=n)`, the range pool now performs on par with (slightly better than) the equivalent tuple pool instead of realizing every draw, and properties over the drawn elements remain solvable.

(should be complementary to https://github.com/pschanely/CrossHair/pull/482 and https://github.com/HypothesisWorks/hypothesis/pull/4802; this is the last of my current set of perf improvements)

### DIFF
--- a/crosshair/libimpl/builtinslib.py
+++ b/crosshair/libimpl/builtinslib.py
@@ -2510,6 +2510,14 @@ class SymbolicRange:
                             raise ValueError
                 self.start, self.stop, self.step = a, b, c
 
+    @classmethod
+    def _from_concrete_range(cls, concrete: range) -> "SymbolicRange":
+        ret = cls.__new__(cls)
+        ret.start = concrete.start
+        ret.stop = concrete.stop
+        ret.step = concrete.step
+        return ret
+
     def __ch_realize__(self):
         start, stop, step = self.start, self.stop, self.step
         return range(realize(start), realize(stop), realize(step))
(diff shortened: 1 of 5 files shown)


pschanely:
Very exciting. I'll review tomorrow!

Zac-HD:
🎊 

Can I suggest also [setting up auto-releases](https://hypothesis.works/articles/continuous-releases/)?  I just tracked down the hypothesis-crosshair clock divergence and realized you fixed it earlier this week, on the weekend I dug into the fpEQ bug and realized you'd already solved that too, etc.  I'm delighted by all this progress and just want to get it into shipped versions sooner!

pschanely:
> 🎊
> 
> Can I suggest also [setting up auto-releases](https://hypothesis.works/articles/continuous-releases/)? I just tracked down the hypothesis-crosshair clock divergence and realized you fixed it earlier this week, on the weekend I dug into the fpEQ bug and realized you'd already solved that too, etc. I'm delighted by all this progress and just want to get it into shipped versions sooner!

I've got releases ~largely automated, but the setup is definitely not as good as yours. (and as you noticed, the human-process side is chaotic) More immediately, I've cut [0.0.109](https://pypi.org/project/crosshair-tool/0.0.109/)!

Zac-HD:
Can we get a `hypothesis-crosshair` release too, with the clock fix?  My CI is going to be _so_ fast when I update...

pschanely:
> Can we get a `hypothesis-crosshair` release too, with the clock fix? My CI is going to be _so_ fast when I update...

Ah of course - [shipped](https://pypi.org/project/hypothesis-crosshair/0.0.29/).
Fingers crossed!

PR #486, opened 2026-07-20
Bind inherited static/class method receivers by method kind

### LINKED ISSUE #451: Inherited (un-overridden) static/class methods mis-bind first argument during class analysis
## Summary

When `analyze_class` analyzes a **concrete class that inherits a `@staticmethod` or `@classmethod` from a base without overriding it**, CrossHair mis-binds the first argument: it generates an *instance* of the subclass and passes it as the method's first positional parameter. This produces spurious `EXEC_ERR` results.

The method's *own* defining class binds correctly — the bug is specific to **inherited, not-overridden** static/class methods surfaced through a subclass.

## Reproduction

```python
import abc

class Base(abc.ABC):
    @abc.abstractmethod
    def area(self) -> int:
        """post: True"""

    @staticmethod
    def snn(n: int) -> bool:
        """post: implies(n >= 0, _)"""
        return n >= 0

    @classmethod
    def lbl(cls) -> str:
        """post: len(_) >= 0"""
        return cls.__name__

class Concrete(Base):
    def area(self) -> int:
        """post: True"""
        return 1
```

```python
from crosshair.core_and_libs import analyze_class, run_checkables
msgs = run_checkables(analyze_class(Concrete))
for m in msgs:
    print(m.state.name, '| line', m.line, '|', m.message[:80])
```

Output:

```
CONFIRMED | line 5  | Confirmed over all paths.
EXEC_ERR  | line 9  | TypeError: '>=' not supported between instances of 'Concrete' and 'int' when calling snn(Concrete(0))
EXEC_ERR  | line 13 | AttributeError: 'Concrete' object has no attribute '__name__' when calling lbl(Concrete(0))
```

Both `snn` (a `staticmethod`) and `lbl` (a `classmethod`) are called with a `Concrete` **instance** as the first argument (`snn(Concrete(0))`, `lbl(Concrete(0))`), i.e. they are analyzed as if they were instance methods.

## Expected

- `snn` should be analyzed with `n: int` (no `self`) → `CONFIRMED`.
- `lbl` should be analyzed with `cls` bound to the class (no instance) → `CONFIRMED`.

Analyzing the base class directly (`analyze_class(Base)`), or defining the same static/class methods directly on the analyzed class, binds correctly — so the defect is in how inherited static/class-method descriptors are resolved for a subclass, most likely in `get_class_conditions`' first-argument handling in `crosshair/condition_parser.py` (the `Conditions.sig` used for argument generation appears to treat the first parameter as `self: Concrete`).

## Notes

- Pre-existing; reproduces independently of the exception-filter work in #449. It was surfaced while writing tests there, and those tests were structured to avoid this shape.
- Environment: observed on Python 3.11.

rahul188:
Closes #451.

### Root cause

In `ConcreteConditionParser.get_class_conditions`, a method picked up from a base
class has its first argument re-typed to the class under analysis:

```python
# Re-type the super's `self` argument to be this class:
revised_sig = set_first_arg_type(super_method_conditions.sig, cls)
```

That is right for instance methods, but it ran for every method kind, and
`set_first_arg_type` just rewrites the annotation of parameter 0. Comparing the
signatures the parser produces for the issue's repro shows what happens:

| method | `Base` | `Concrete` (before) |
|---|---|---|
| `area` (instance) | `(self: Base)` | `(self: Concrete)` — correct |
| `lbl` (classmethod) | `(cls: Type[Base])` | `(cls: Concrete)` — lost `Type[...]` |
| `snn` (staticmethod) | `(n: int)` | `(n: Concrete)` — clobbered a real parameter |

So the staticmethod's own leading parameter was overwritten, and the classmethod
was handed an *instance* where it expected the class — hence
`snn(Concrete())` / `lbl(Concrete())` and the spurious `EXEC_ERR`s. The base class
analyzed fine because its methods come from `cls.__dict__` and never take this path.

A related crash falls out of the same line: an inherited staticmethod with no
parameters made `set_first_arg_type` index an empty list, raising
`IndexError: list index out of range` out of `analyze_class`.

### Fix

Branch on the descriptor kind (`inspect.getattr_static`) before rebinding:
classmethods get `Type[cls]`, staticmethods are left alone, instance methods keep
the existing behavior.

### Testing

The issue's repro now reports `CONFIRMED` for all four methods instead of two
`EXEC_ERR`s, and `Concrete`'s signatures mirror `Base`'s with the class substituted:

```
area: (self: Concrete) -> int
lbl:  (cls: Type[Concrete]) -> str
snn:  (n: int) -> bool
```

Added two tests to `condition_parser_test.py` asserting the first argument for each
of the three method kinds, and covering the zero-parameter inherited staticmethod.
Both fail on `main` (the second with the `IndexError`) and pass with the change.

`crosshair/{core,fnutil,condition_parser,main}_test.py`: 159 passed, 3 skipped.
Smoke suite: 13 passed. black, isort, and flake8 clean.

Changelog entry added under *Next Version*.

### DIFF
--- a/crosshair/condition_parser.py
+++ b/crosshair/condition_parser.py
@@ -486,8 +486,18 @@ class ConcreteConditionParser(ConditionParser):
             method = cls.__dict__.get(method_name, None)
             super_method_conditions = super_methods.get(method_name)
             if super_method_conditions is not None:
-                # Re-type the super's `self` argument to be this class:
-                revised_sig = set_first_arg_type(super_method_conditions.sig, cls)
+                # Re-type the super's receiver argument to be this class.
+                # A staticmethod has no receiver; its first argument is an
+                # ordinary parameter and is left alone.
+                descriptor = inspect.getattr_static(cls, method_name, None)
+                if isinstance(descriptor, staticmethod):
+                    revised_sig = super_method_conditions.sig
+                elif isinstance(descriptor, classmethod):
+                    revised_sig = set_first_arg_type(
+                        super_method_conditions.sig, Type[cls]
+                    )
+                else:
+                    revised_sig = set_first_arg_type(super_method_conditions.sig, cls)
                 super_method_conditions = replace(
                     super_method_conditions, sig=revised_sig
                 )
--- a/crosshair/condition_parser_test.py
+++ b/crosshair/condition_parser_test.py
@@ -1,7 +1,7 @@
 import inspect
 import json
 import sys
-from typing import List
+from typing import List, Type
 
 import pytest
 
@@ -111,6 +111,31 @@ class SubClassExample(BaseClassExample):
         return 5
 
 
+class ReceiverKindsExample:
+    def instance_method(self) -> int:
+        """post: True"""
+        return 1
+
+    @staticmethod
+    def static_method(n: int) -> bool:
+        """post: implies(n >= 0, _)"""
+        return n >= 0
+
+    @classmethod
+    def class_method(cls) -> str:
+        """post: len(_) >= 0"""
+        return cls.__name__
+
+    @staticmethod
+    def static_method_without_args() -> bool:
+        """post: _"""
+        return True
+
+
+class InheritsReceiverKinds(ReceiverKindsExample):
+    pass
+
+
 def test_parse_sections_variants() -> None:
     parsed = parse_sections([(1, " :post: True ")], ("post",), "")
     assert set(parsed.sections.keys()) == {"post"}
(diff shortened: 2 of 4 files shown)


pschanely (approved):
This is so great: thank you!
Care to add yourself to the Contributor List in `contributing.rst` before I merge? (and link to whatever you want)

rahul188:
Thank you! Added myself to the contributor list in `contributing.rst` (linked to my GitHub). Ready whenever you are.

PR #489, opened 2026-07-21
Model read-only symbolic memoryviews

### LINKED ISSUE #445: SymbolicMemoryView is always writable — memoryview readonly not modeled
## Summary
`SymbolicMemoryView` never models a read-only memoryview. `register_type(memoryview, lambda p: SymbolicMemoryView(p(bytearray)))` always backs the view with a mutable `SymbolicByteArray`, and `SymbolicMemoryView.__init__` sets `self.readonly = isinstance(obj, bytes)` — which is therefore always `False`. So a symbolic memoryview is always writable, whereas a concrete `memoryview(b"...")` is read-only.

## Impact
Write-through operations diverge from CPython:
- Concrete: `memoryview(b"ab")[0] = 1` → `TypeError: cannot modify read-only memory`
- Symbolic: succeeds (mutates the backing bytearray)

This can produce false divergences in the single-operation differential (`fuzz_core_test` / `tools/measure_support`) for memoryview write ops, and generally under-models read-only views.

## Where
- `crosshair/libimpl/builtinslib.py` — `SymbolicMemoryView.__init__`: `self.readonly = isinstance(obj, bytes)`
- `crosshair/libimpl/builtinslib.py` — `register_type(memoryview, lambda p: SymbolicMemoryView(p(bytearray)))`

The `__setitem__` guard already raises `TypeError` when `self.readonly` — the flag is just never `True`.

## Related
`SymbolicMemoryView` intentionally has no `_smt_for_unification` (a view's equality is only over the viewed bytes, no clean full SMT form), so memoryview inputs currently can't be pinned in the differential and are skipped. Modeling `readonly` is likely a prerequisite to meaningfully enabling memoryview coverage there.

## Possible directions
- Have `proxy_for_type(memoryview)` sometimes choose a bytes (read-only) backing and set `readonly` accordingly.
- Ensure write ops honor `readonly`.

Found while adding the differential support-measurement (branch `demo-support-measurement`).

w3lld1:
## Summary

- I made symbolic `memoryview` inputs choose between `bytes` and `bytearray` backing so both read-only and writable views are reachable.
- I added regression coverage showing that writes to a generated read-only view raise `TypeError`.
- I updated the existing `toreadonly()` test and changelog for the new input model.

## Validation

- `PYTHONHASHSEED=0 python -m pytest -q crosshair/libimpl/builtinslib_test.py` (514 passed, 2 skipped)
- `PYTHONHASHSEED=0 python -m pytest -q crosshair/libimpl/builtinslib_ch_test.py -k 'memoryview or buffer_setitem'` (5 passed)
- `black --check crosshair/libimpl/builtinslib.py crosshair/libimpl/builtinslib_test.py`
- `isort --check-only crosshair/libimpl/builtinslib.py crosshair/libimpl/builtinslib_test.py`
- `flake8 crosshair/libimpl/builtinslib.py crosshair/libimpl/builtinslib_test.py`
- `python -m doctest doc/source/changelog.rst`
- `git diff --check`

The full cross-platform suite is left to CI.

Fixes #445

### DIFF
--- a/crosshair/libimpl/builtinslib.py
+++ b/crosshair/libimpl/builtinslib.py
@@ -5426,7 +5426,10 @@ def make_registrations():
     # Text: (elsewhere - identical to str)
     register_type(bytes, make_byte_string)
     register_type(bytearray, lambda p: SymbolicByteArray(p(bytes)))
-    register_type(memoryview, lambda p: SymbolicMemoryView(p(bytearray)))
+    register_type(
+        memoryview,
+        lambda p: SymbolicMemoryView(make_union_choice(p, bytes, bytearray)),
+    )
     # AnyStr,  (it's a type var)
 
     register_type(typing.BinaryIO, lambda p: io.BytesIO(p(bytes)))
--- a/crosshair/libimpl/builtinslib_test.py
+++ b/crosshair/libimpl/builtinslib_test.py
@@ -4127,10 +4127,23 @@ def test_memoryview_toreadonly():
         mv = proxy_for_type(memoryview, "mv")
         space.add(mv.__len__() == 1)
         mv2 = mv.toreadonly()
-        mv[0] = 12
-        assert mv2[0] == 12
+        if not mv.readonly:
+            mv[0] = 12
+            assert mv2[0] == 12
         with pytest.raises(TypeError):
             mv2[0] = 24
+
+
+def test_memoryview_can_be_readonly():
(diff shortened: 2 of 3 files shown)


pschanely:
@w3lld1 Thank you so much for doing this! It looks like someone else picked the same issue in #490 and added a few bonus fixes, so I ended up merging that one. Thanks again, and I hope to hear from you again on github, discord, etc!

PR #490, opened 2026-07-21
Model read-only memoryviews

### LINKED ISSUE #445: SymbolicMemoryView is always writable — memoryview readonly not modeled

rahul188:
Fixes #445.

## The problem
`register_type(memoryview, ...)` always backed a symbolic memoryview with a mutable `SymbolicByteArray`, so `readonly` was never `True` and writes always succeeded — where a concrete `memoryview(b"...")` raises `TypeError`. The `__setitem__` guard was already correct; the flag just had no way to become `True` through `proxy_for_type`.

## The fix
- `proxy_for_type(memoryview)` now forks between a `bytes` (read-only) and `bytearray` (writable) backing (`make_memoryview`, following the `datetimelib` `smt_fork` registration pattern). `__init__`'s existing `isinstance(obj, bytes)` derivation does the rest under tracing.
- While testing the flag I found two adjacent leaks and fixed them too:
  - **Slicing dropped `readonly`** — and it turned out the symbolic slice branch in `__getitem__` was dead code: it checked `isinstance(newslice, SliceView)` under tracing, where `SliceView` reports its emulated type (`tuple`), so every unit-step slice silently took the `realize` path. The check now runs under `NoTracing`, and the new view inherits the parent's `readonly` (matching CPython, where a slice of a `toreadonly()` view stays read-only).
  - **Realization dropped `readonly`** — `__ch_realize__`/`__ch_deep_realize__` rebuilt from the backing object, so realizing a `toreadonly()` view over a bytearray produced a writable concrete view. Both now apply `.toreadonly()` when the symbolic view is read-only.

## Verification
- New `test_proxied_memoryview_can_be_readonly` (`post: not mv.readonly` → `POST_FAIL`) fails on `main` (CrossHair confirms the postcondition because no readonly view exists) and passes with the fix; `test_proxied_memoryview_can_be_writable` guards the other side of the fork.
- New slice assertion in `test_memoryview_toreadonly` fails on `main` (slice came back writable) and passes with the fix. Both failures were observed before applying the source change.
- `test_memoryview_toreadonly` now builds its view over an explicit `bytearray` proxy and `test_memoryview_properties` compares against a backing-matched concrete view, so both are deterministic under the new fork.
- Full suite (`CROSSHAIR_EXTRA_ASSERTS=1 PYTHONHASHSEED=0 pytest -n auto --dist worksteal crosshair`): 14193 passed; the handful of failures (demo_overrides encoding, datetime astimezone, one flaky decimal case) fail identically on unmodified `main` in this Windows environment.
- black / isort / flake8 / mypy: no new findings.

### DIFF
--- a/crosshair/libimpl/builtinslib.py
+++ b/crosshair/libimpl/builtinslib.py
@@ -4511,13 +4511,15 @@ class SymbolicMemoryView(BytesLike):
         sliced = self._sliced
         obj, start, stop = self.obj, sliced.start, sliced.stop
         self.obj = obj
-        return memoryview(realize(obj))[realize(start) : realize(stop)]
+        ret = memoryview(realize(obj))[realize(start) : realize(stop)]
+        return ret.toreadonly() if self.readonly else ret
 
     def __ch_deep_realize__(self, memo):
         sliced = self._sliced
         obj, start, stop = self.obj, sliced.start, sliced.stop
         self.obj = obj
-        return memoryview(deep_realize(obj, memo))[realize(start) : realize(stop)]
+        ret = memoryview(deep_realize(obj, memo))[realize(start) : realize(stop)]
+        return ret.toreadonly() if self.readonly else ret
 
     def __ch_pytype__(self):
         return memoryview
@@ -4551,14 +4553,14 @@ class SymbolicMemoryView(BytesLike):
     def __getitem__(self, key):
         if isinstance(key, slice):
             newslice = self._sliced[key]
-            if isinstance(newslice, SliceView):
-                with NoTracing():
+            with NoTracing():
+                if isinstance(newslice, SliceView):
                     ret = SymbolicMemoryView(self.obj)
                     ret._sliced = newslice
+                    ret.readonly = self.readonly
                     return ret
-            else:
-                # Give up when there's a step in the slice:
-                return realize(self).__getitem__(key)
+            # Give up when there's a step in the slice:
+            return realize(self).__getitem__(key)
         else:
             return self._sliced[key]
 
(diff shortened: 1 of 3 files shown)


pschanely:
Amazing - thank you!!

PR #492, opened 2026-07-22
Realize z3 Reals as representable floats

### LINKED ISSUE #491: 0.0.109 regression: symbolic float can realize to a non-IEEE-representable rational, breaking path constraints on conversion
**Versions:** regressed in crosshair-tool 0.0.109 (0.0.108 is fine); observed via hypothesis-crosshair 0.0.28 and 0.0.29, CPython 3.12 & 3.14, Linux.

## Summary

Under the hypothesis-crosshair provider, a `RealBasedSymbolicFloat` whose path constraints include `x != 1` (plus finiteness/ordering constraints against another symbolic float) gets *realized* to the rational `(N+1)/N` with `N ≈ 2e320` — i.e. `1 + 2**-1064`-ish, a real number which is **not representable as an IEEE double**. Converting that symbolic to a concrete `float` (via `float(x)` / C-level `nb_float` extraction) then rounds to exactly `1.0`, which violates the `x != 1` path constraint. Downstream code that compares `float(x) == x` observes `False` for what should be the same number, and Hypothesis raises a spurious `InvalidArgument: max_value=1.0 cannot be exactly represented as a float of width 64 - use max_value=1.0 instead.`

Probing at the point of failure (inside `st.floats()` validation, with `NoTracing`):

```
raw type(max_arg)=RealBasedSymbolicFloat   raw type(max_value)=float
max_arg=1.0  max_value=1.0                 # reprs — both realize/print as 1.0
var=float_03
space.is_possible(float_03 == RealVal(1.0)) -> False   # !!
space.is_possible(float_03 != RealVal(1.0)) -> True
assertions mentioning float_03:
    float_03_isfinite_3
    0 < float_03
    float_03 > float_05
    1 != float_03
    (200...001/200...000) == float_03      # realization pinned to 1 + (tiny)
```

i.e. the model value chosen at realization time satisfies the Real-arithmetic constraints but is not a float64; every subsequent conversion to a concrete float silently lands on the *excluded* value `1.0`.

## Reproducer

Deterministic with the Hypothesis test suite (master, or any recent release):

```bash
pip install "crosshair-tool==0.0.109" "hypothesis-crosshair==0.0.29" pytest
cd hypothesis/hypothesis
HYPOTHESIS_PROFILE=crosshair python -m pytest \
    "tests/cover/test_float_nastiness.py::test_fuzzing_floats_bounds[nonfloat1]"
# -> InvalidArgument / FlakyBackendFailure, every run
# with crosshair-tool==0.0.108 the same command passes every run
```

The failing test draws two symbolic floats `lo`/`hi`, swaps them if `lo > hi`, and passes them as bounds to `st.floats(min_value=lo, max_value=hi, width=64)`, whose validation does `float_of(hi, 64) != hi` (a concrete-vs-symbolic comparison after an f64 extraction).

I could not trigger the bad model with a small standalone `standalone_statespace` script (z3 picks `2.0` for simple `x > 0 ∧ x != 1` constraint sets), so the full assertion set above — in particular the interaction with a second symbolic float and the finiteness/smallest-magnitude encoding — seems to be needed to steer z3 into the `1 + epsilon` corner.

## Expected

Realization of a symbolic float should only ever produce values that are exactly representable in the modeled float type, so that later `float(x)`-style conversions cannot contradict path constraints.

## Notes

Presumably related to the 0.0.109 changes to float equality encoding ("Fix symbolic floats treating `0` and `-0.0` as unequal" — the `!=` constraints now reaching z3 in a different form), though the mechanism (non-representable model value at realization) may well predate 0.0.109 and merely be newly-reachable.

pschanely:
Fixes #491

### DIFF
--- a/crosshair/statespace.py
+++ b/crosshair/statespace.py
@@ -4,11 +4,13 @@ import contextvars
 import copy
 import enum
 import functools
+import math
 import random
 import re
 import threading
 from collections import Counter, defaultdict
 from dataclasses import dataclass
+from fractions import Fraction
 from sys import _getframe
 from time import process_time
 from traceback import extract_stack, format_tb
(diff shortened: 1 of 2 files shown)


pschanely:
Abandoned for the moment.
Details here: https://github.com/pschanely/CrossHair/issues/491#issuecomment-5052471712

PR #512, opened 2026-09-19
Reduce the fixed per-path cost of installing patches and interceptors

KRRT7:
## Problem

Every execution path enters `Patched()`, which did two things that scale badly with the number of paths explored:

- `PatchingModule.add` rebuilt the function override table from all ~390 registrations, entry by entry in Python, and `pop` unwound it the same way. Measured at ~885µs per path.
- It pushed the ten opcode interceptor modules one at a time, and `CompositeTracer.push_module` called `sys.monitoring.restart_events()` for each. Under real instrumentation a restart costs ~50µs, so ~500µs per path.

Together that's roughly 1.5ms of fixed overhead per path, which profiles as 5-12% of wall time on workloads that explore many short paths.

## Change

- `PatchingModule` caches each applied override layer, keyed on the registrations dict and the base state it was applied to. A hit is validated by dict equality (values are functions, so this compares by identity and is cheap), which means registrations added later still invalidate the cache. `pop` restores the saved base state directly. The cache is bounded at 16 entries so transient dicts (e.g. `EvalFriendlyReprContext`) can't grow it.
- `CompositeTracer.push_modules()` installs a batch of tracing modules with a single event restart. `Patched.__enter__` and `is_deeply_immutable` use it. The restart now happens after the push rather than before; the tracer is disabled while these run, so no events can be disabled in between either way.

## Results

Per path: patch install+remove 885µs → 5µs; event restarts 10 → 1.

`crosshair-benchmark` suite, 60s cap, median of 3 runs, two independent runs of this branch against one baseline run:

| Benchmark group | Change |
|---|---|
| Sum of medians, all 32 | −3% to −5% |
| Four unsolved regex benchmarks (timeout-bound) | ~0% (expected) |
| Solved multi-second runs (`find_5_times`, `missing_nonzero_exp`, `missing_exponent`, `exp_in_upper_or_lower_case`, `compare_version_strings`, `missing_negative_zero`) | −8% to −22%, consistent across both runs |
| Sub-second runs | within process-startup noise (±20%) |

## Testing

- `tracers_test`, `util_test`, `enforce_test`, `core_test`, `main_test`, `opcode_intercept_test`, `copyext_test`, `condition_parser_test`, `libimpl/builtinslib_test`: 717 passed, 2 skipped. Smoke suite passes.
- black, isort, flake8 clean on the changed files.


pschanely:
@KRRT7 Awesome performance debugging and diagnosis!

After further review, I ended up taking a slightly different approach and lifted the patching so that it happens per-check rather than per-path: landed in https://github.com/pschanely/CrossHair/pull/520 (note that perf there was run against harder benchmark challenges - improvement %s are not comparable)

